PR #2, opened 2015-07-29
ios-log and ios-crash-log first run through

syang97:


lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
+import path from 'path';
+import _fs from 'fs';
+import _ from 'lodash';
+import glob from 'glob';
+import { getLogger } from 'appium-support';
+import B from 'bluebird';
+
+let fs = {
```
sebv: fs is in appium-support already, probably just add missing method there.

jlipps: yeah both of these are now in appium-support, so you can just `import { fs } from 'appium-support'`, and then use `await fs.stat` etc as necessary.

also a note about this code: you're promisifying the sync version of stat, which probably won't do what you had in mind!

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+class IosCrashLog { 
+  constructor() {
+    this.prevLogs = [];
+    this.logsSinceLastRequest = [];
+    this.crashDir = path.resolve(process.env.HOME, 'Library', 'Logs', 'DiagnosticReports');
+  }
+
+  async getCrashes() {
+    try {
+      let crashFiles = await glob(`${this.crashDir} /*.crash`);
```
jlipps: in this case we can just `return await glob...` to avoid creating an unnecessary variable

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+      return crashFiles;
+    } catch (err) {
+      logger.error('There was a problem getting the crash list');
```
jlipps: we can use `logger.errAndThrow` for this case

jlipps: but we might want to just avoid the try/catch here and let the original error bubble up. what do you think @imurchie?

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+let logger = getLogger('ios-crash-log');
+
+class IosCrashLog { 
+  constructor() {
+    this.prevLogs = [];
+    this.logsSinceLastRequest = [];
+    this.crashDir = path.resolve(process.env.HOME, 'Library', 'Logs', 'DiagnosticReports');
+  }
+
+  async getCrashes() {
```
jlipps: `async getCrashes ()`

(i.e., add a space before the open paren; check all the function definitions for this)

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+      return crashFiles;
+    } catch (err) {
+      logger.error('There was a problem getting the crash list');
+      throw err;
+    }
+  }
+
+  async filesToJSON(arr) {
+    let logs = [];
+    for (let i = 0; i < arr.length; i++) {
```
jlipps: when iterating we prefer to use `let...of` nowadays. I'd change `arr` to `files` and then do:

```js
for (let file of files) {
```

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+    return logs;
+  }
+
+  async startCapture() {
+    let fileList = await this.getCrashes();
+    this.prevLogs = fileList;
+  }
+
+  getLogs() {
+    let crashFiles = glob.sync(`${this.crashDir} /*.crash`);
```
jlipps: let's not use sync methods

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
+import path from 'path';
+import _fs from 'fs';
+import _ from 'lodash';
+import glob from 'glob';
```
jlipps: you'll probably need to promisify this if you want to use it with `await`

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+  async startCapture() {
+    let fileList = await this.getCrashes();
+    this.prevLogs = fileList;
+  }
+
+  getLogs() {
+    let crashFiles = glob.sync(`${this.crashDir} /*.crash`);
+    let diff = _.difference(crashFiles, this.prevLogs, this.logsSinceLastRequest);
+    this.logsSinceLastRequest = _.union(this.logsSinceLastRequest, diff);
+    return this.filesToJSON(diff);
```
jlipps: `filesToJSON` is an async method, which means `getLogs` needs to be an async method as well

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  getAllLogs() {
```
jlipps: same thing about being an async method here

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+    let crashFiles = glob.sync(`${this.crashDir} /*.crash`);
+    let logFiles = _.difference(crashFiles, this.prevLogs);
+    return this.filesToJSON(logFiles);
+  }
+}
+
+export function IosCrashLog() {
```
jlipps: this function has the same name as the class, which I don't think will work. also, the function creates a new object but doesn't do anything with it.

personally I think we should just `export default IosCrashLog;` and be done with it. the user can call `new` just fine.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+import _ from 'lodash';
+import glob from 'glob';
+import { getLogger } from 'appium-support';
+import B from 'bluebird';
+
+let fs = {
+  stat: B.promisify(_fs.statSync),
+  readFile: B.promisify(_fs.readFile)
+};
+require('date-utils');
```
imurchie: Remaining `require`.

lib/ios-crash-log.js
imurchie: We also don't need the `await` with the `return`. The `async` function will just wrap the value from the `await` in a promise, so you might as well just return the promise from `glob` (once it is promisified).

lib/ios-crash-log.js
imurchie: I think the former is good. We get logging attached to where it is thrown, but the original error bubbles up.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,68 @@
...
+let logger = getLogger('ios-crash-log');
```
imurchie: We need to standardize the naming scheme. Below is `iOS-log`, here `ios-crash-log`.

jlipps: FWIW I've been using UpperCamelCase for log names, like BaseDriver, AndroidDriver, etc... If we want to follow that, these should be IOSLog and IOSCrashLog

lib/ios-log.js
```diff
@@ -0,0 +1,237 @@
...
+let START_TIMEOUT = 10000;
+let logger = getLogger('iOS-log');
+
+class IosLog { 
+  constructor(opts) {
+    this.udid = opts.udid;
+    this.simUdid = opts.simUdid;
+    this.showLogs = opts.showLogs;
+    this.proc = null;
+    this.onIosLogStart = null;
```
imurchie: It feels like this callback should be subsumed in the starting promise.

lib/ios-log.js
```diff
@@ -0,0 +1,235 @@
...
+  async startCapture () {
+    if (this.udid) { 
+      this.loggingModeOn = false;
+      let spawnEnv = _.clone(process.env);
+      logger.debug('Attempting iOS device log capture via libimobiledevice idevicesyslog');
+      let err = await which('idevicesyslog');
+      if (!err) { 
+        this.proc = spawn('idevicesyslog', {env: spawnEnv});
+      } else {
+        logger.warn('Could not capture device log using libimobiledevice idevicesyslog.' +
```
jlipps: this might be a complicated change, but we probably want to use `SubProcess` instead of `spawn`, since it it's more 'async/await' friendly

lib/ios-log.js
```diff
@@ -0,0 +1,235 @@
...
+          if (_.isUndefined(this.simUdid)) {
+            logger.errorAndThrow('ios8 log capture requires a sim udid'); 
+          }
+          logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'CoreSimulator', this.simUdid); 
+        } else {
+          logger.debug('Starting iOS7.* simulator log capture');
+          logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+        }
+
+        let touchLog = async function () {
```
jlipps: part of the reason we created all of these intermediate functions (touchLog, getLogFiles, etc...), was because in ES5 callback land, it made things more readable. now with async/await, it actually makes it harder to understand. personally I think we might want to do away with the intermediate functions and first see how it looks just with all the code in one long block.

if it ends up being too long we can always break it up later, but then it will be according to rationale that makes sense for our new situation.

lib/ios-log.js
```diff
@@ -0,0 +1,235 @@
...
+        };
+        try {
+          await touchLog();
+          await getLogFiles();
+          await tailLogs();
+        } catch (err) {
+          logger.warn('System log capture failed.');
+          return;
+        }
+      } else {
```
jlipps: here's a pattern that I like to use when I have long conditionals is. the way the code is currently structured (someone did this before you got on the scene, of course), is:

```js
if (this.udid) {
  // many lines of code
} else {
  if (ver >= 5) {
    // many many lines of code
  } else {
    // 3 lines of code
  }
}
```

This gets hard to read because there are so many lines of code it can be easy to get lost and know which branch of the conditional you're in. We also lose horizontal space due to indentations. A way to solve that is:

```js
if (this.udid) {
  // many lines of code
  return;
}

if (ver < 5) {
  // 3 lines of code
  return;
}

// many lines of code
```

This "short-circuiting" kind of pattern can definitely be used to add confusion instead of take it away, but in this case I think it might make things more clear. What do you think @imurchie?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+      let spawnEnv = _.clone(process.env);
+      logger.debug('Attempting iOS device log capture via libimobiledevice idevicesyslog');
+      let err = await which('idevicesyslog');
+      if (!err) { 
+        this.proc = new SubProcess('idevicesyslog', {env: spawnEnv});
+      } else {
+        logger.warn('Could not capture device log using libimobiledevice idevicesyslog.' +
+            ' Libimobiledevice probably is not installed');
+        logger.debug('Attempting iOS device log capture via deviceconsole');
+        let limDir = path.resolve(__dirname, '../../../build/deviceconsole');  //???
```
moizjv: //??? ?

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
+import path from 'path';
+
+import _ from 'lodash';
+import _glob from 'glob';
+import { fs, getLogger } from 'appium-support';
+import B from 'bluebird';
+let glob = B.promisify(_glob);
```
jlipps: i like this kind of thing to be a `const`

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+      logger.debug('Starting iOS6.* simulator log capture');
+      this.proc = new SubProcess('tail', ['-f', '-n', '1', '/var/log/system.log']);
+      this.finishStartingLogCapture();
+      return;
+    }
+
+    let logsPath;
+
+    if (ver >= 6) { 
+      logger.debug('Starting iOS 8.* simulator log capture');
```
moizjv: We would soon release 9.* support to so can we make this iOS 8+ or some thing which indicates iOS 8 and iOS 9. What do you think @jlipps ?

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+
+
+require('date-utils');
+let logger = getLogger('IOS-crash-log');
```
jlipps: same here

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+        if (_.isUndefined(this.simUdid)) {
+          logger.errorAndThrow('ios8 log capture requires a sim udid');
```
moizjv: same as above

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+
+class IOSCrashLog { 
+  constructor() {
+    this.prevLogs = [];
+    this.logsSinceLastRequest = [];
+    this.crashDir = path.resolve(process.env.HOME, 'Library', 'Logs', 'DiagnosticReports');
```
jlipps: personally, since this is just a global thing (unrelated to the state of the object), i'd make it a const up top, like:

```
const CRASH_DIR = ...;
```

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+    } else {
+        logger.debug('Starting iOS7.* simulator log capture');
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+    }
+
+    try {
+      if (logsPath.indexOf('*') >= 0) {
+        return; 
+      }
+        await mkdirp(logsPath);
```
moizjv: indentation

lib/ios-log.js
```diff
@@ -0,0 +1,208 @@
...
```
imurchie: Indentation.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  async getCrashes () {
+    try {
+      return glob(`${this.crashDir} /*.crash`);
```
jlipps: this is kind of a tricky case. the `try/catch` will only work if we actually `await`, so in this case, it does really need to be `return await glob(...);`

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+        await fs.appendFile(path.resolve(logsPath, 'system.log'), 
+              'A new Simulator is about to start!'); 
+        let files;
+        try {
+          files = await glob(path.resolve(logsPath, 'system.log'));
+        if (files.length < 1) {
+          throw 'Could not start log capture';
+        }
+      } catch (e) {
+        logger.error('Could not start log capture because no iOS ' +
```
moizjv: We can use new interpolation here

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+    } catch (err) {
+      logger.errorAndThrow('There was a problem getting the crash list');
```
jlipps: `errorAndThrow` already throws, so no need to `throw err` again.

also, in cases like these, it's usually helpful to include the original error along with the nice message, like:

```
logger.errorAndThrow(`There was a problem getting the crash list: ${err}`);
```

That way we have at least something to go on in debugging

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+        'simulator logs could be found at ' + logsPath + '/system.log' +
+        'Logging will not be functional for this run');
+      }
+
+      let lastModifiedLogPath = files[0];
+      let lastModifiedLogTime = await fs.stat(files[0]).mtime; //promisify this
```
moizjv: good comment :+1:

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+    if (!this.proc) {
+      let msg = 'Could not capture device log';
+      logger.warn(msg);
+      throw new Error(msg);
+    }
+     this.startTimer = setTimeout(function () {
+      let msg = 'Log capture did not start in a reasonable amount of time';
+      logger.error(msg);
+
+      throw new Error(msg);
```
moizjv: why not `errorAndThrow` here?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+  // 
+
+  // first time you want to ignore it, 
+  // have the promise use a local varialbe outside of promise, and change the boolean after the first line is done. 
+  // make first lien empty 
+  finishStartingLogCapture () {
+    if (!this.proc) {
+      let msg = 'Could not capture device log';
+      logger.warn(msg);
+      throw new Error(msg);
```
moizjv: `errorAndThrow` here too?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+      logger.error(msg);
+
+      throw new Error(msg);
+      
+    }, START_TIMEOUT);
+    let firstLine = true;
+    return new Promise((resolve, reject) => {
+      this.proc.on('output', (stdout, stderr) => {
+        if (stdout) {
+          //log.info(`[STDOUT] ${stdout.trim()}`);
```
moizjv: ?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+          clearTimeout(this.startTimer);
+          if (firstLine) {
+            this.onOutput(stdout);
+          } else {
+            this.logRow += stdout;
```
moizjv: Will this code be ever called because i think `firstLine` would never be set to `false`, am i missing something?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+            if (stdout.substr(-1,1) === '\n') {
+              this.onOutput(stdout, '');
+              this.logRow = ''; 
+              firstLine = false;
+            }
+          }         
+          resolve();
+        }
+        if (stderr) {
+          //log.info(`[STDERR] ${stderr.trim()}`);
```
moizjv: ?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+          clearTimeout(this.startTimer);
+          if(/execvp\(\)/.test(stderr)) {
+            logger.error('iOS log capture process failed to start');
+            reject(new Error('iOS log capture process failed to start'));
+            return;
```
moizjv: Do we need return here?

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+            }
+          reject(stderr);
+          this.onOutput(stderr, ' STDERR');
```
moizjv: I am not sure if this will print after rejection? @jlipps what do you think, in theory it should print but i am not sure

lib/ios-log.js
syang97: oops, put the line in the wrong block!

lib/ios-log.js
```diff
@@ -0,0 +1,217 @@
...
+
+  stopCapture () {
+    logger.debug('Stopping iOS log capture');
+    if (this.proc) {
+      this.proc.kill();
+    }
+    this.proc = null; 
+  }
+
+  onOutput (data, prefix) {
```
moizjv: where and how are we using `data`?

lib/ios-log.js
moizjv: @syang97 so `onOutput` should come before rejecting it

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+      throw err;
+    }
+  }
+
+  async filesToJSON (files) {
+    let logs = [];
```
jlipps: this could be more elegantly written as a map, maybe?

```js
import { asyncmap } from 'asyncbox';
...
return asyncmap(files, async (f) => {
  let filename = path.resolve(this.crashDir, file);
  let stat = await fs.stat(filename);
  return {
    timestamp: stat.ctime,
    level: 'ALL', 
    message: await fs.readFile(filename, 'utf8')
  };
});
```

lib/ios-crash-log.js
```diff
@@ -0,0 +1,63 @@
...
+    for (let file of files) {
+      let filename = path.resolve(this.crashDir, file);
+      let stat = await fs.stat(filename);
+      let logObj = {
+        timestamp: stat.ctime,
+        level: 'ALL', 
+        message: fs.readFile(filename, 'utf8')
```
jlipps: pretty sure this needs to be `await`ed

lib/ios-log.js
```diff
@@ -0,0 +1,208 @@
...
+      let lastModifiedLogTime = await fs.stat(files[0]).mtime;
+      for (let file in files) {
+        let mtime = await fs.stat(file).mtime;
+        if (mtime > lastModifiedLogTime) {
+          lastModifiedLogPath = file;
+          lastModifiedLogTime = mtime;
+        }
+      }
+      this.proc = new SubProcess('tail', ['-f', '-n', '1', lastModifiedLogPath]);
+      this.finishStartingLogCapture();
```
jlipps: all these calls to `this.finishStartingLogCapture()` should probably be `await`ed, right? we don't actually want to be done with `startCapture` until that's done too?

lib/ios-log.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+     this.startTimer = setTimeout(function () {
+      let msg = 'Log capture did not start in a reasonable amount of time';
+      logger.error(msg);
+
+      throw new Error(msg);
+      
+    }, START_TIMEOUT);
+    let firstLine = true;
+    return new Promise((resolve, reject) => {
```
jlipps: seeing this line makes me think that you're not using the `SubProcess` API correctly; docs on it are here: https://github.com/appium/node-teen_process#teen_processsubprocess

let's touch base (with me or @imurchie) tomorrow to go over how it works in more detail

syang97: Ah okay, I didn't see that documentation.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,61 @@
...
+    this.prevLogs = [];
+    this.logsSinceLastRequest = [];
+  }
+
+  async getCrashes () {
+    try {
+      return await glob(`${CRASH_DIR} /*.crash`);
+    } catch (err) {
+      logger.errorAndThrow(`There was a problem getting the crash list: ${err}`);
+      throw err;
```
jlipps: `errorAndThrow` will throw, so this line will never be reached and can be got rid of

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+import { SubProcess } from 'teen_process';
+
+// Date-Utils: Polyfills for the Date object
+
+let START_TIMEOUT = 10000;
+let logger = getLogger('IOS-log');
+
+class IOSLog { 
+
+  constructor(opts) {
```
jlipps: let's go through a list of these fields and make sure they are actually used, otherwise let's get rid of unused ones

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      if (!err) { 
+        this.proc = new SubProcess('idevicesyslog', {env: spawnEnv});
+      } else {
+        logger.warn('Could not capture device log using libimobiledevice idevicesyslog.' +
+            ' Libimobiledevice probably is not installed');
+        logger.debug('Attempting iOS device log capture via deviceconsole');
+        let limDir = path.resolve(__dirname, '../../../build/deviceconsole');
+        spawnEnv.PATH = `${process.env.PATH}:${limDir}`;
+        spawnEnv.DYLD_LIBRARY_PATH = `${limDir}:${process.env.DYLD_LIBRARY_PATH}`;
+        this.proc =  new SubProcess('deviceconsole', ['-u', this.udid], {env: spawnEnv});
```
jlipps: looks like an extra space after the = here

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+    this.iosLogStartTime = null;
+    this.loggingModeOn = true;
+    this.logs = [];
+    this.logRow = '';
+    this.logsSinceLastRequest = [];
+    this.startTimer = null;
+  }
+
+  async startCapture () {
+    if (this.udid) {
```
jlipps: let's add a comment: `// if we have a real device...'`

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        logger.debug('Attempting iOS device log capture via deviceconsole');
+        let limDir = path.resolve(__dirname, '../../../build/deviceconsole');
+        spawnEnv.PATH = `${process.env.PATH}:${limDir}`;
+        spawnEnv.DYLD_LIBRARY_PATH = `${limDir}:${process.env.DYLD_LIBRARY_PATH}`;
+        this.proc =  new SubProcess('deviceconsole', ['-u', this.udid], {env: spawnEnv});
+      }
+      await this.finishStartingLogCapture();
+      return;
+    } 
+
```
jlipps: now here we can add another comment: `// otherwise, if we have a simulator...`

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+    let xcodeVersion = await xcode.getVersion();
+    let ver = parseInt(xcodeVersion.split('.')[0], 10);
+
+    if ( ver < 5 ) {
```
jlipps: don't need spaces around the condition

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        }
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'CoreSimulator', this.simUdid); 
+    } else {
+        logger.debug('Starting iOS7.* simulator log capture');
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+    }
+
+    try {
+      if (logsPath.indexOf('*') >= 0) {
+        return;
```
jlipps: i'm curious why we are returning here if `logsPath` has a `*` in it. this has the effect of not erroring, but likewise we are not starting any log capture...

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        logger.debug('Starting iOS7.* simulator log capture');
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+    }
+
+    try {
+      if (logsPath.indexOf('*') >= 0) {
+        return; 
+      }
+      await mkdirp(logsPath);
+      await touch(path.resolve(logsPath, 'system.log'));
```
jlipps: since we make the same call to path.resolve here, let's do:

```js
let systemLogPath = path.resolve(logsPath, 'system.log');
```

then we can reuse it

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      await fs.appendFile(path.resolve(logsPath, 'system.log'), 
+            'A new Simulator is about to start!'); 
+      let files;
+      try {
+        files = await glob(path.resolve(logsPath, 'system.log'));
```
jlipps: can also then use it here :-)

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        if (files.length < 1) {
+          throw 'Could not start log capture';
```
jlipps: should be:
```js
throw new Error("Could not start log capture");
```

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        }
+      } catch (e) {
+        logger.error(`Could not start log capture because no iOS ` +
+        `simulator logs could be found at ${logsPath}/system.log` +
```
jlipps: we line subsequent string lines up under the beginning of the first line's string. does that make sense? kinda hard to explain in words...

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        let mtime = await fs.stat(file).mtime;
+        if (mtime > lastModifiedLogTime) {
+          lastModifiedLogPath = file;
+          lastModifiedLogTime = mtime;
+        }
+      }
+      this.proc = new SubProcess('tail', ['-f', '-n', '1', lastModifiedLogPath]);
+      await this.finishStartingLogCapture();
+    } catch (err) {
+      logger.warn('System log capture failed.');
```
jlipps: I guess I think that we should `errorAndThrow` this instead; and let the calling method ignore errors if they exist

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      return;
```
jlipps: this `return` is the last thing in the method and therefore does nothing. let's get rid of it

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+  async finishStartingLogCapture () {
+    if (!this.proc) {
+      logger.errorAndThrow('Could not capture device log');
+    }
+    this.startTimer = setTimeout(function () {
+      logger.errorAndThrow('Log capture did not start in a reasonable amount of time');
+    }, START_TIMEOUT);
+    let firstLine = true;
+
+    //this.proc.on --> start --> function startDetector
```
jlipps: we can get rid of this comment

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+    this.proc.on('output', (stdout, stderr) => {
+      if (stdout) {
+        if (firstLine) {
+          if (stdout.substr(-1,1) === '\n') {
+              this.onOutput();
```
jlipps: looks like these lines are indented too much

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+    this.startTimer = setTimeout(function () {
+      logger.errorAndThrow('Log capture did not start in a reasonable amount of time');
+    }, START_TIMEOUT);
+    let firstLine = true;
+
+    //this.proc.on --> start --> function startDetector
+    this.proc.on('output', (stdout, stderr) => {
+      if (stdout) {
+        if (firstLine) {
+          if (stdout.substr(-1,1) === '\n') {
```
jlipps: `(-1, 1)`

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+              this.onOutput();
+              firstLine = false; 
+          } 
+        } else {
+          this.logRow += stdout;
+          if (stdout.substr(-1,1) === '\n') {
```
jlipps: `(-1, 1)`

lib/ios-log.js
imurchie: Now `this.onIosLogStart` is not used at all.

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+            this.logRow = ''; 
+          }
+        }
+      } 
+      if (stderr) {
+        this.onOutput(' STDERR');
+      }
+    });
+
+    let sd = (stdout, stderr) => {
```
jlipps: pretty sure the startDetector also needs to return `true` whenever we actually do have a successful start condition. in our case, that would probably be anytime we have any output, so you can do:

```js
return stdout || stderr;
```

also, it should read:

```js
if (/execvp...
```

(space after `if`)

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      if(/execvp\(\)/.test(stderr)) {
+        throw new Error('iOS log capture process failed to start');      
+      }
+    };
+
+    await this.proc.start(sd, START_TIMEOUT);
```
jlipps: nice! this was much cleaner, wasn't it?

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        if (!this.loggingModeOn) {
+          let logRowParts = log.split(/\s+/);
+          let logRowDate = new Date(Date.parse(`${this.iosLogStartTime.getFullYear()} ${logRowParts[0]} ${logRowParts[1]} ${logRowParts[2]}`));
+            if (logRowDate.isAfter(this.iosLogStartTime)) {
+            this.loggingModeOn = true;
+            }
+          }
+        if (this.loggingModeOn) {
+          let logObj = {
+            timeStamp: Date.now()
```
jlipps: we do comma-last style now

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+            , level: 'ALL'
+            , message: log
+          };
+          this.logs.push(logObj);
+          this.logsSinceLastRequest.push(logObj);
+            if (this.showLogs) logger.info(`[IOS_SYSLOG_ROW ${prefix}] ${log}`);
```
jlipps: indentation; also we try to avoid one-line ifs in most situations, so let's add some braces and make it a normal-looking 3-line if statement

lib/ios-crash-log.js
```diff
@@ -0,0 +1,61 @@
...
+import _ from 'lodash';
+import _glob from 'glob';
+import { fs, getLogger } from 'appium-support';
+import B from 'bluebird';
+import { asyncmap } from 'asyncbox';
+
+const glob = B.promisify(_glob);
+
+
+require('date-utils');
```
imurchie: I don't see where we use the `date-utils` polyfills in this file.

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+    } else {
+        logger.debug('Starting iOS7.* simulator log capture');
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+    }
+
+    try {
+      if (logsPath.indexOf('*') >= 0) {
+        return; 
+      }
+      await mkdirp(logsPath);
```
imurchie: Does `mkdirp` not need to be promisified?

lib/ios-log.js
imurchie: Or split it into a `startCaptureRealDevice` method.

lib/ios-crash-log.js
jlipps: @imurchie it essentially monkeypatches `Date` when it's required; we should add a comment to explain why this isn't a normal require

lib/ios-log.js
jlipps: :+1: to that idea of having multiple methods, probably cleaner to read

lib/ios-log.js
imurchie:
```js
logger.error(`Could not start log capture because no iOS ` +
             `simulator logs could be found at ${logsPath}/system.log` +
             `Logging will not be functional for this run`);
```

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      this.iosLogStartTime = new Date();
+    }
+
+    let logs = this.logRow.split('\n');
+    for (let log in logs) {
+      log = log.trim();
+      if (log) {
+        if (!this.loggingModeOn) {
+          let logRowParts = log.split(/\s+/);
+          let logRowDate = new Date(Date.parse(`${this.iosLogStartTime.getFullYear()} ${logRowParts[0]} ${logRowParts[1]} ${logRowParts[2]}`));
```
imurchie: Does not `new Date()` implicitly call `Date.parse()`?

lib/ios-crash-log.js
imurchie: I think it should be moved to the other file, as well, since (a) we already have a comment there, and (b) we seem to actually use it there (I don't see any of the monkeypatched `Date` methods in use in this file).

lib/ios-crash-log.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+  }
+
+  async startCapture () {
+    let fileList = await this.getCrashes();
+    this.prevLogs = fileList;
+  }
+
+  async getLogs () {
+    let crashFiles = await glob(`${CRASH_DIR} /*.crash`);
```
imurchie: Do we really want a space in this path? It looks weird. If it **is** supposed to be there, a little note might be in order so that it doesn't catch the eye.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,61 @@
...
+    let diff = _.difference(crashFiles, this.prevLogs, this.logsSinceLastRequest);
+    this.logsSinceLastRequest = _.union(this.logsSinceLastRequest, diff);
+    return this.filesToJSON(diff);
+  }
+
+  async getAllLogs() {
+    let crashFiles = await glob(`${CRASH_DIR} /*.crash`);
```
imurchie: Ditto space here. Since we have it in multiple places, and it is static, the path might also be better placed in a constant like `CRASH_FILES` or somesuch.

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      this.proc = new SubProcess('tail', ['-f', '-n', '1', '/var/log/system.log']);
+      await this.finishStartingLogCapture();
+      return;
+    }
+
+    let logsPath;
+
+    if (ver >= 6) { 
+      logger.debug('Starting iOS 8.*/9.* simulator log capture');
+        if (_.isUndefined(this.simUdid)) {
```
imurchie: Indentation.

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+          logger.errorAndThrow('iOS 8*/9.* log capture requires a sim udid'); 
+        }
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'CoreSimulator', this.simUdid);
```
imurchie: This might not be the place for this suggestion to come in, but it seems like all of this information ought not be bleeding out into the rest of Appium. Rather than being passed a `simUdid`, we ought to be passed a simulator object, from which we get information of this sort.

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+        logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'iOS Simulator', '7.*');
+    }
+
+    try {
+      if (logsPath.indexOf('*') >= 0) {
+        return; 
+      }
+      await mkdirp(logsPath);
+      await touch(path.resolve(logsPath, 'system.log'));
+      await fs.appendFile(path.resolve(logsPath, 'system.log'),
```
imurchie: Do we really need to `touch` the file before appending? `fs.appendFile()` [should](https://nodejs.org/api/fs.html#fs_fs_appendfile_filename_data_options_callback) create the file if it doesn't already exist. Or is there some other reason for doing the `touch`?

lib/ios-log.js
```diff
@@ -0,0 +1,203 @@
...
+      } catch (e) {
+        logger.error(`Could not start log capture because no iOS ` +
+        `simulator logs could be found at ${logsPath}/system.log` +
+        `Logging will not be functional for this run`);
+      }
+
+      let lastModifiedLogPath = files[0];
+      let lastModifiedLogTime = await fs.stat(files[0]).mtime;
+      for (let file in files) {
+        let mtime = await fs.stat(file).mtime;
```
imurchie: I don't think `stat` is part of the `appium-support` `fs` module.

jlipps: i think it is; i and sebv added a bunch of fs bits to appium-support

lib/ios-log.js
jlipps: this is a good suggestion; but for now, should we put a // TODO to this effect so we know how to refactor it in a later pass?

lib/ios-crash-log.js
jlipps: yeah, i'd be surprised if a space is intentional here.

of course at some point we'll want a bunch of functional tests which should hopefully catch these kinds of issues

lib/ios-log.js
```diff
@@ -0,0 +1,205 @@
...
+    let logsPath;
+
+    if (ver >= 6) { 
+      logger.debug('Starting iOS 8.*/9.* simulator log capture');
+      if (_.isUndefined(this.simUdid)) {
+        logger.errorAndThrow('iOS 8*/9.* log capture requires a sim udid'); 
+      }
+      logsPath = path.resolve(process.env.HOME, 'Library', 'Logs', 'CoreSimulator', this.simUdid); 
+    } else {
+        logger.debug('Starting iOS7.* simulator log capture');
```
imurchie: Indentation.

lib/ios-crash-log.js
```diff
@@ -0,0 +1,59 @@
...
+    try {
+      return await glob(`${CRASH_DIR}`);
+    } catch (err) {
+      logger.errorAndThrow(`There was a problem getting the crash list: ${err}`);
+    }
+  }
+
+  async filesToJSON (files) {
+    return asyncmap(files, async (f) => {
+      let filename = path.resolve(this.crashDir, f);
```
imurchie: Where is `this.crashDir` initialized?

lib/ios-crash-log.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+  }
+
+  async startCapture () {
+    let fileList = await this.getCrashes();
+    this.prevLogs = fileList;
+  }
+
+  async getLogs () {
+    let crashFiles = await glob(`${CRASH_DIR}`);
```
imurchie: Is there a reason to do it manually, and not use `this.getCrashes()`?

lib/ios-crash-log.js
```diff
@@ -0,0 +1,59 @@
...
+
+class IOSCrashLog { 
+  constructor() {
+    this.prevLogs = [];
+    this.logsSinceLastRequest = [];
+  }
+
+  async getCrashes () {
+    try {
+      return await glob(`${CRASH_DIR}`);
```
imurchie: No need to interpolate `CRASH_DIR

lib/ios-crash-log.js
```diff
@@ -0,0 +1,59 @@
...
+
+  async getLogs () {
+    let crashFiles = await glob(`${CRASH_DIR}`);
+    let diff = _.difference(crashFiles, this.prevLogs, this.logsSinceLastRequest);
+    this.logsSinceLastRequest = _.union(this.logsSinceLastRequest, diff);
+    return this.filesToJSON(diff);
+  }
+
+  async getAllLogs() {
+    let crashFiles = await glob(`${CRASH_DIR}`);
```
imurchie: Ditto manually getting the files.

lib/ios-log.js
```diff
@@ -0,0 +1,202 @@
...
+      spawnEnv.PATH = `${process.env.PATH}:${limDir}`;
+      spawnEnv.DYLD_LIBRARY_PATH = `${limDir}:${process.env.DYLD_LIBRARY_PATH}`;
+      this.proc = new SubProcess('deviceconsole', ['-u', this.udid], {env: spawnEnv});
+    }
+    await this.finishStartingLogCapture();
+  }
+
+  async startCapture () {
+    if (this.udid) { // if we have a real device
+      await this.startCaptureRealDevice();
```
imurchie: Do you want to do the rest of this process when it is a real device? If not, you need to `return this.startCaptureRealDevice();`.

lib/ios-crash-log.js
jlipps: agreed, we shouldn't reimplement the same logic, let's use our helper function since it exists

lib/ios-log.js
```diff
@@ -0,0 +1,202 @@
...
+import touch from 'touch';
+import xcode from 'xcode'; 
+import B from 'bluebird';
+import { SubProcess } from 'teen_process';
+
+// Date-Utils: Polyfills for the Date object
+require('date-utils'); 
+
+let START_TIMEOUT = 10000;
+let logger = getLogger('IOS-log');
```
jlipps: still think we should call this logger 'iOSLog'

lib/ios-crash-log.js
```diff
@@ -0,0 +1,59 @@
...
+import _ from 'lodash';
+import _glob from 'glob';
+import { fs, getLogger } from 'appium-support';
+import B from 'bluebird';
+import { asyncmap } from 'asyncbox';
+
+const glob = B.promisify(_glob);
+
+
+const logger = getLogger('IOS-crash-log');
```
jlipps: think we should call this logger 'iOSCrashLog', also no need for extra spaces above it

lib/ios-log.js
```diff
@@ -0,0 +1,202 @@
...
+import _mkdirp from 'mkdirp';
+import touch from 'touch';
+import xcode from 'xcode'; 
+import B from 'bluebird';
+import { SubProcess } from 'teen_process';
+
+// Date-Utils: Polyfills for the Date object
+require('date-utils'); 
+
+let START_TIMEOUT = 10000;
```
jlipps: these `let`s should all be `const`s

lib/ios-log.js
imurchie: Added. https://github.com/appium/appium-support/pull/17

jlipps: ok fine :-) can you cut a new version of appium-support?

lib/ios-log.js
jlipps: @soniayang97 looks like you missed this suggestion perhaps?

PR #4, opened 2015-08-13
Wire up the logs and add some basic tests

imurchie:
Fix some issues with Sonia's logs, and add a couple of tests to make sure the basic features are functional.


lib/ios-log.js
```diff
@@ -51,55 +58,47 @@ class IOSLog {
...
       return;
-    }
-
-    let logsPath;
-
-    if (ver >= 6) {
-      logger.debug('Starting iOS 8.*/9.* simulator log capture');
-      if (_.isUndefined(this.simUdid)) {
-        logger.errorAndThrow('iOS 8*/9.* log capture requires a sim udid');
+    } else if (_.isUndefined(this.simUdid)) {
```
jlipps: i think if we have a `return` it can just be `if` instead of `else if`; the latter makes me feel like there's some important reason to have `else` when there's not.

lib/logger.js
```diff
@@ -1,5 +1,5 @@
 import { getLogger } from 'appium-logger';
 
-let logger = getLogger('iosCrashLog');
+let logger = getLogger('iOS');
```
jlipps: hmm, we've been following a convention of having a different log name per package, partially to make it easier to figure out where to grep for log messages if something bad happens.

test/unit/log-specs.js
```diff
@@ -0,0 +1,109 @@
...
+import { getSimulator } from 'appium-ios-simulator';
+import ncp from 'ncp';
+import sinon from 'sinon';
+import { fs } from 'appium-support';
+import B from 'bluebird';
+import path from 'path';
+import logger from '../../lib/logger';
+
+
+let copy = B.promisify(ncp.ncp);
```
jlipps: i put `copyFile` on appium-support/fs (though sebv didn't like it much)

lib/logger.js
imurchie: But this is the `appium-ios-driver` package. I'm not sure how `iosCrashLog` came to be in there. I would be happy to break this out into its own package, but I'm not sure if we use it for anything else (though, come to think of it, it could be useful for some other packages, like `appium-ios-simulator`, and `appium-uiauto`, to have access to the logs).

test/unit/log-specs.js
imurchie: Cool!

imurchie:
I decided that the iOS logging is generally useful in some other packages (e.g., `appium-ios-simulator`), so I broke it out into its own package, https://github.com/appium/appium-ios-log

What remains of this PR is some cleanup.

lib/logger.js
jlipps: good point, I forgot this wasn't its own package.

and now it is! looks like I was not forgetting but rather remembering the future :-)

PR #5, opened 2015-08-14
Add desired capabilities constraints

imurchie:
Add the constraints to iOS desired capabilities, and wire into the driver. See [here](https://github.com/appium/appium-base-driver/blob/master/lib/desired-caps.js) for basis of validation.

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -1,0 +1,86 @@
+
+
+const desiredCapConstraints = {
+  platformName: {
+    presence: true,
+    isString: true,
+    inclusion: ['iOS']
+  },
+  browserName: {
+    isString: true,
+    inclusion: ['Safari']
+  },
+  app: {
+    isString: true
+  },
+
+  calendarFormat: {
+    isString: true
+  },
+  bundleId: {
+    isString: true
+  },
+  udid: {
+    isString: true
+  },
+  locationServicesEnabled: {
+    isBoolean: true
+  },
+  locationServicesAuthorized: {
+    isBoolean: true
+  },
+  autoAcceptAlerts: {
+    isBoolean: true
+  },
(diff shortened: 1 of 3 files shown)


lib/desired-caps.js
```diff
@@ -0,0 +1,87 @@
...
+  platformName: {
+    presence: true,
+    isString: true,
+    inclusion: ['iOS']
+  },
+  browserName: {
+    isString: true,
+    inclusion: ['Safari']
+  },
+  app: {
```
jlipps: we'll also probably want to override the validating function to perform meta-cap validation, e.g., that either `bundleId` or `app` must exist

lib/desired-caps.js
```diff
@@ -0,0 +1,87 @@
+const desiredCapConstraints = {
+  platformName: {
+    presence: true,
+    isString: true,
+    inclusion: ['iOS']
```
jlipps: is this an ios-driver level constraint? or is this an appium-level constraint?

imurchie: My understanding is that it is an Appium-level constraint, but we override it here to specify that the `platformName` can only be `iOS`.

jlipps: ah. do we want to require presence? i guess this is a broader question of whether this driver should require the full gamut of mjsonwp caps, or whether it should assume that it can only do iOS, thus the cap isn't really required...

imurchie: Good point. We will only instantiate the iOS driver if the `platformName` is `iOS`, so there is no real need for this check. I'm not sure if it would be best to have it anyway, as documentation of the specific caps (since technically it _is_ required).

jlipps: probably better to be strict. let's go with that.

PR #6, opened 2015-08-19
Remove settings code

imurchie:
Use the simulator object instead.


sebv:
Cool lgtm.

lib/driver.js
```diff
@@ -389,7 +359,7 @@ class IosDriver extends BaseDriver {
       locServ = locServ || this.capabilities.locationServicesAuthorized;
       locServ = locServ ? 1 : 0;
       logger.debug("Setting location services to " + locServ);
-      settings.updateSettings(this.sim, 'locationServices', {
+      this.sim.updateSettings('locationServices', {
```
jlipps: i'm surprised that we are not `await`ing these; hopefully there's not still sync code in here somewhere?

PR #8, opened 2015-08-20
Add log commands

imurchie:
Commands structure, and first couple of commands.

### DIFF
--- a/lib/commands/logging.js
+++ b/lib/commands/logging.js
@@ -1,0 +1,33 @@
+import _ from 'lodash';
+import logger from '../logger';
+
+let commands = {};
+
+const SUPPORTED_LOG_TYPES = {
+  'syslog': 'System Logs - Device logs for iOS applications on real devices and simulators',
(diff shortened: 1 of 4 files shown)


sebv:
lgtm

lib/commands/logging.js
```diff
@@ -0,0 +1,40 @@
...
+
+let commands = {};
+
+const SUPPORTED_LOG_TYPES = {
+  'syslog': 'System Logs - Device logs for iOS applications on real devices and simulators',
+  'crashlog': 'Crash Logs - Crash reports for iOS applications on real devices and simulators',
+  'performance': 'Performance Logs - Debug Timelines on real devices and simulators'
+};
+
+commands.getLogTypes = function () {
```
jlipps: as with my comments on the other repo, I think all commands should be decorated with `async` because that is how MJSONWP is going to interact with them.

imurchie: That sounds good. I was moving in that direction as I worked on other commands, too.

PR #14, opened 2015-08-26
Add basic context handling and web commands

imurchie:
Passes all the iOS webview tests, not Safari tests.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -1,0 +1,458 @@
+import _ from 'lodash';
+import B from 'bluebird';
+import { retryInterval } from 'asyncbox';
+import { RemoteDebugger, WebKitRemoteDebugger } from 'appium-remote-debugger';
+import { IOSPerformanceLog } from 'appium-ios-log';
+import { errors } from 'mobile-json-wire-protocol';
+import logger from '../logger';
+
+
+const NATIVE_WIN = 'NATIVE_APP';
+const WEBVIEW_WIN = 'WEBVIEW';
+const WEBVIEW_BASE = `${WEBVIEW_WIN}_`;
+
+let commands = {}, helpers = {}, extensions = {};
+
+commands.getCurrentContext = async function () {
+  if (this.curContext && this.curContext !== NATIVE_WIN) {
+    return `${WEBVIEW_BASE}${this.curContext}`;
+  } else {
+    return NATIVE_WIN;
+  }
+};
+
+commands.getContexts = async function () {
+  logger.debug('Getting list of available contexts');
+  let contexts = await this.getContextsAndViews();
+  return contexts.map((context) => context.id);
+};
+
+commands.setContext = async function (name, callback, skipReadyCheck) {
+  function alreadyInContext (desired, current) {
+    return (desired === current ||
+           (desired === null && current === NATIVE_WIN) ||
+           (desired === NATIVE_WIN && current === null));
+  }
+
+  logger.debug(`Attempting to set context to '${name}'`);
+  if (alreadyInContext(name, this.curContext)) {
+    // already in the named context, no need to do anything
+  } else if (name === NATIVE_WIN || name === null) {
+    // switching into the native context
+    this.curContext = null;
+    if (this.isRealDevice()) {
+      this.remote.disconnect();
+    }
+  } else {
+    // switching into a webview context
+    let idx = name.replace(WEBVIEW_BASE, '');
+    if (idx === WEBVIEW_WIN) {
+      // allow user to pass in "WEBVIEW" without an index
+      idx = '1';
+    }
+
+    // if contexts have not already been retrieved, get them
+    if (_.isUndefined(this.contexts)) {
+      await this.getContexts();
+    }
+
+    if (!_.contains(this.contexts, idx)) {
+      throw new errors.NoSuchContextError();
+    }
+    let pageIdKey = parseInt(idx, 10);
+
+    if (!this.isRealDevice()) {
+      await this.remote.selectPage(pageIdKey, skipReadyCheck);
+      this.curContext = idx;
+    } else {
+      if (this.remote) {
+        await this.remote.disconnect();
+      }
+      this.curContext = idx;
+      await this.remote.connect(idx);
+    }
+  }
+
+  // attempt to start performance logging, if requested
+  if (this.perfLogEnabled && this.remote) {
+    logger.debug(`Starting performance log on '${this.curContext}'`);
+    this.logs.performance = new IOSPerformanceLog(this.remote);
+    this.logs.performance.startCapture();
+  }
+};
+
+commands.getWindowHandle = async function () {
+  if (!this.isWebContext()) {
+    throw new errors.NotImplementedError();
+  }
+  return this.curContext;
+};
+
+commands.getWindowHandles = async function () {
+  if (!this.isWebContext()) {
+    throw new errors.NotImplementedError();
+  }
+
+  let pageArray = await this.listWebFrames();
+  this.windowHandleCache = _.map(pageArray, this.massagePage);
+  let idArray = _.pluck(this.windowHandleCache, 'id');
+  // since we use this.contexts to manage selecting debugger pages, make
+  // sure it gets populated even if someone did not use the
+  // getContexts method
+  if (!this.contexts) {
+    this.contexts = idArray;
+  }
+  return idArray;
+};
+
+commands.setWindowHandle = async function (name, skipReadyCheck) {
+  if (!this.isWebContext()) {
+    throw new errors.NotImplementedError();
+  }
+
+  if (!_.contains(_.pluck(this.windowHandleCache, 'id'), name)) {
+    throw new errors.NoSuchWindowError();
+  }
+  let pageIdKey = parseInt(name, 10);
+  if (!this.isRealDevice()) {
+    await this.remote.selectPage(pageIdKey, skipReadyCheck);
+    this.curContext = this.curWindowHandle = name;
+  } else {
+    if (name === this.curWindowHandle) {
+      logger.debug(`Remote debugger is already connected to window '${name}'`);
+    } else if (!_.contains(_.pluck(this.windowHandleCache, 'id'), name)) {
+      throw new errors.NoSuchWindowError();
+    } else {
+      await this.remote.disconnect();
+      this.curContext = this.curWindowHandle = name;
+      await this.remote.connect(name);
+    }
+  }
+};
+
+extensions.initAutoWebview = async function () {
+  if (this.caps.autoWebview) {
+    logger.debug('Setting auto webview');
+    await this.navToInitialWebview(this);
+  }
+};
+
+extensions.getContextsAndViews = async function () {
+  logger.debug('Retrieving contexts and views');
+  let webviews = await this.listWebFrames();
+  let ctxs = [{id: NATIVE_WIN}];
+  this.contexts = [NATIVE_WIN];
+  for (let view of webviews) {
+    ctxs.push({id: `${WEBVIEW_BASE}${view.id}`, view});
+    this.contexts.push(view.id.toString());
+  }
+  return ctxs;
+};
+
+extensions.listWebFrames = async function () {
+  if (!this.caps.bundleId) {
+    logger.errorAndThrow('Cannot enter web frame without a bundle ID');
+  }
+
+  let pageArray;
+  if (this.remote !== null && this.caps.bundleId !== null) {
+    if (this.isRealDevice()) {
+      pageArray = await this.remote.pageArrayFromJson();
+    } else {
+      pageArray = await this.remote.selectApp(this.caps.webviewConnectRetries);
+    }
+  } else {
+    if (this.isRealDevice()) {
+      this.remote = new WebKitRemoteDebugger();
+      return this.remote.pageArrayFromJson();
+    }
+    this.remote = new RemoteDebugger({
+      bundleId: this.caps.bundleId,
+      useNewSafari: this.useNewSafari(),
+      pageLoadMs: this.pageLoadMs,
+      platformVersion: this.caps.platformVersion
+    });
+
+    let appInfo = await this.remote.connect();
+    if (!appInfo) {
+      logger.debug('Unable to connect to the remote debugger.');
+      this.remote = null;
+      return [];
+    }
+    pageArray = await this.remote.selectApp(this.caps.webviewConnectRetries);
+    this.remote.on(RemoteDebugger.EVENT_PAGE_CHANGE, this.onPageChange.bind(this));
+
+    let tryClosingAlert = async () => {
+      let didDismiss = await this.closeAlertBeforeTest();
+      if (!didDismiss) {
+        throw new Error('Close alert failed. Retry.');
+      }
+    };
+    try {
+      await retryInterval(3, 4000, tryClosingAlert);
+    } catch (err) {
+      // if the loop to close alerts failed to dismiss, ignore,
+      // otherwise log and throw the error
+      if (err.message !== 'Close alert failed. Retry.') {
+        logger.errorAndThrow(err);
+      }
+    }
+  }
+
+  if (pageArray.length === 0) {
+    // we have no web frames, so disconnect from the remote debugger
+    this.stopRemote();
+  }
+  return pageArray;
+};
+
+extensions.onPageChange = async function (pageArray) {
+  logger.debug(`Remote debugger notified us of a new page listing: ${JSON.stringify(pageArray)}`);
+  if (this.selectingNewPage) {
+    logger.debug('We are in the middle of selecting a page, ignoring');
+    return;
+  }
+  let newIds = [];
+  let newPages = [];
+  let keyId = null;
+  for (let page of pageArray) {
+    let id = page.id.toString();
+    newIds.push(id);
+    if (page.isKey) {
+      keyId = id;
+    }
+    if (!_.contains(this.contexts, id)) {
+      newPages.push(id);
+      this.contexts.push(id);
+    }
+  }
+
+  let newPage = null;
+  if (this.curContext === null) {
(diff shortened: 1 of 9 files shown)


lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+};
+
+commands.getContexts = async function () {
+  logger.debug('Getting list of available contexts');
+  let contexts = await this.getContextsAndViews();
+  return contexts.map((context) => context.id);
+};
+
+commands.setContext = async function (name, callback, skipReadyCheck) {
+  function alreadyInContext(desired, current) {
```
jlipps: `function alreadyInContext (desired, current) {`

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+
+    if (_.contains(this.contexts, idx)) {
+      let pageIdKey = parseInt(idx, 10);
+
+      this.processingRemoteCmd = true;
+      if (!this.isRealDevice()) {
+        await this.remote.selectPage(pageIdKey, skipReadyCheck);
+        this.curContext = idx;
+      } else {
+        if (name === this.curContext) {
```
jlipps: yeah isn't this our first check above, at the beginning of the function?

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+      //     this.closeAlertBeforeTest(function (err, didDismiss) {
+      //       if (!didDismiss) {
+      //         setTimeout(loopClose, 1000);
+      //       }
+      //     });
+      //   }
+      // }.bind(this);
+      // setTimeout(loopClose, 4000);
+      await B.delay(4000);
+      // TODO: figure out how to do the alert stuff
```
jlipps: yess....

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+    } else {
+      logger.error('Do not have our current window anymore, and there ' +
+                   'are not any more to load! Doing nothing...');
+      return;
+    }
+    this.curContext = keyId;
+    // this.remote.pageIdKey = parseInt(keyId, 10);
+    newPage = keyId;
+  } else {
+    let dirty = () => {
```
jlipps: if we just call this one time, let's make it a self-executing anonymous function so that `dirty` is just a bool. this also deserves a comment.

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+      var item = (arr) => {
+        return _.filter(arr, function (obj) {
+          return obj.id === this.curContext;
+        }, this)[0];
+      };
+
+      return !_.isEqual(item(this.contexts), item(pageArray));
+    };
+
+    // If a window gets navigated to an anchor it doesn't always fire a page callback event
```
jlipps: or maybe just moving the comment up

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+  let matchingCtx;
+  // _(contexts).each((ctx) => {
+  for (let ctx of contexts) {
+    if (ctx.view && (ctx.view.title || '').match(titleRegex)) {
+      if (ctx.view.url === 'about:blank') {
+        // in the case of Xcode  < 5 (i.e., iOS SDK Version less than 7)
+        // and in the case of iOS 7.1 in a webview (not in Safari)
+        // we can have the url be `about:blank`
+        if (parseFloat(this.iOSSDKVersion) < 7 ||
+            (this.caps.platformVersion === '7.1' && this.caps.app && this.caps.app.toLowerCase() !== 'safari')) {
```
jlipps: we should be using `this.opts`

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+  return matchingCtx ? matchingCtx.id : undefined;
+};
+
+// Right now we don't necessarily wait for webview
+// and frame to load, which leads to race conditions and flakiness
+// , let see if we can transition to something better
+extensions.useNewSafari = function () {
+  return parseFloat(this.iosSdkVersion) >= 8.1 &&
+         parseFloat(this.caps.platformVersion) >= 8.1 &&
+         !this.caps.udid &&
```
jlipps: same here, probably want to do a global find/re

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+
+extensions.typeAndNavToUrl = async function () {
+  let initialUrl = this.caps.safariInitialUrl || `http://127.0.0.1:${this.caps.port}/welcome`;
+  let oldImpWait = this.implicitWaitMs;
+  this.implicitWaitMs = 7000;
+  let res = await this.findElement('name', 'URL');
+  this.implicitWaitMs = oldImpWait;
+  await this.nativeTap(res.value.ELEMENT);
+  res = await this.findElements('name', 'Address');
+  let addressEl = res.value[res.value.length -1].ELEMENT;
```
jlipps: space

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+          }
+        } catch (err) {
+          logger.warn('Could not click reload button, continuing');
+          await B.delay(spinTime);
+          return await spinHandles();
+        }
+      } else {
+        throw new Error('Could not navigate to webview; there are none!');
+      }
+    } else {
```
jlipps: this should probably go up first under an `if (res)` so it can short-circuit and we need fewer indentation levels

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+      res = await this.getLatestWebviewContextForTitle(titleRegex);
+    } catch (err) {
+      throw new Error(`Could not navigate to webview! Err: ${err.message}`);
+    }
+    if (!res) {
+      if ((Date.now() - start) < 90000) {
+        logger.warn("Could not find any webviews yet, refreshing/retrying");
+        if (this.caps.udid || !this.caps.safari) {
+          await B.delay(spinTime);
+          return await spinHandles();
```
jlipps: don't understand why we would call spinHandles again just after waiting. if that's all we want to do, why not just wait at the top of spinHandles?

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+        }
+        try {
+          res = await this.findUIElementOrElements('accessibility id', 'ReloadButton', '', false);
```
jlipps: surprised that this doesn't just return the value directly; i thought we'd moved away from internal functions knowing anything about jsonwp response structure

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+      await this.remote.cancelPageLoad();
+      return;
+    }
+  };
+  await spinHandles();
+};
+
+helpers.closeAlertBeforeTest = async function () {
+  let res = await this.uiAutoClient.sendCommand('au.alertIsPresent()');
+  if (res !== null && !_.isUndefined(res) && res.value === true) {
```
jlipps: same q about value here

lib/commands/web.js
```diff
@@ -0,0 +1,193 @@
...
+    this.processingRemoteCmd = false;
+    return value;
+  } else {
+    throw new errors.NotImplementedError();
+  }
+};
+
+commands.title = async function () {
+  if (this.isWebContext()) {
+    let res =  await this.executeAtom('title', []);
```
jlipps: space

lib/commands/web.js
```diff
@@ -0,0 +1,193 @@
...
+
+  return res;
+};
+
+helpers.lookForAlert = async function (counter) {
+  if (!_.isNull(this.instruments) && !_.isNull(this.uiAutoClient)) {
+    if (!this.returnedFromExecuteAtom[counter] && !this.selectingNewPage) {
+      logger.debug('atom did not return yet, checking to see if ' +
+                   'we are blocked by an alert');
+      // temporarily act like we're not processing a remote command
```
jlipps: is this comment still relevant? i don't see us setting any processingRemoteCommand variable here

jlipps:
ok, done with first pass!

lib/commands/context.js
sebv: We don't want to wait the first time.

lib/commands/context.js
sebv: it does, can be renamed el.

lib/commands/context.js
sebv: yeah, just following my bad example, we need to change them in all the files in the repo.

lib/commands/context.js
```diff
@@ -0,0 +1,472 @@
...
+extensions.navToInitialWebview = async function () {
+  let timeout = 0;
+  if (this.isRealDevice()) {
+    timeout = 3000;
+    logger.debug(`Waiting for ${timeout} ms before navigating to view.`);
+  }
+  await B.delay(timeout);
+  if (this.useNewSafari()) {
+    await this.typeAndNavToUrl();
+  } else if (parseInt(this.iosSdkVersion, 10) >= 7 && !this.caps.udid && this.caps.safari) {
```
sebv: Let's create some `isRealDevice` and `isSafari` functions.

lib/commands/web.js
```diff
@@ -0,0 +1,193 @@
...
+  let res;
+  let doFind = async () => {
+    res = await this.executeAtom(`find_element${many ? 's' : ''}`, [strategy, selector, atomsElement]);
+    return res.status === 0 && res.value !== null;
+  };
+  await this.implicitWaitForCondition(doFind);
+  return res.value;
+};
+
+commands.click = async function (elementId) {
```
sebv: This is going to conflict with the gesture one. This probably need to be a `webClick` helper.

sebv:
Let's merge it quick so we get both native/web test working from master before Friday.

lib/commands/context.js
imurchie: There is an `isRealDevice` (Added on line 501 of device.js).

sebv: Then we should try to use them instead of `this.caps.udid` or `this.caps.safari`. I know is not done yet in a lot of places on master.

lib/commands/context.js
imurchie: Yes, I'll table this as a to-do.

imurchie:
Addressed the issues. Some things are not yet tested since we only use them in Safari. We will go through and get rid of `this.caps` at a later time.

Depends, now, on https://github.com/appium/appium-remote-debugger/pull/4

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+  if (this.useNewSafari()) {
+    await this.typeAndNavToUrl();
+  } else if (parseInt(this.iosSdkVersion, 10) >= 7 && !this.isRealDevice() && this.caps.safari) {
+    await this.navToViewThroughFavorites();
+  } else {
+    await this.navToViewWithTitle(/.*/);
+  }
+};
+
+extensions.typeAndNavToUrl = async function () {
```
jlipps: this function is so much easier to read now

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+    let msg = 'Could not find button to click to get into webview. ' +
+              'Proceeding on the assumption we have a working one.';
+    logger.error(msg);
+    this.implicitWaitMs = oldImpWait;
+    return await this.navToViewWithTitle(/.*/i);
+  }
+  this.implicitWaitMs = oldImpWait;
+  try {
+    let res = await this.nativeTap(el.ELEMENT);
+    if (res.status !== 0) {
```
jlipps: is this another erroneous assumption about receiving a jsonwp response instead of an actual value? can we not get rid of this try/catch and simply allow `nativeTap` to throw if it doesn't work?

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+    } catch (err) {
+      throw new Error(`Could not navigate to webview! Err: ${err.message}`);
+    }
+    if (res) {
+      let latestWindow = res;
+      logger.debug(`Picking webview '${latestWindow}'`);
+      await this.setContext(latestWindow);
+      await this.remote.cancelPageLoad();
+      return;
+    } else {
```
jlipps: since we have a return here, no need for the `else`. just dedent the following.

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+      if ((Date.now() - start) < 90000) {
```
jlipps: I think it would be more elegant if there were one try/catch in here and one `return await spinHandles()`

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+          return await spinHandles();
+        }
+        try {
+          await this.nativeTap(element.ELEMENT);
+        } catch (err) {
+          logger.warn('Could not click reload button, continuing');
+          await B.delay(spinTime);
+          return await spinHandles();
+        }
+      } else {
```
jlipps: like a previous comment i made, since this is a tiny little block that ends control of this function, why don't we reverse the boolean and put this up top? then we need fewer indented brace blocks :-)

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+};
+
+helpers.closeAlertBeforeTest = async function () {
+  let present = await this.uiAutoClient.sendCommand('au.alertIsPresent()');
+  if (present) {
+    logger.debug('Alert present before starting test, let us banish it');
+    await this.uiAutoClient.sendCommand('au.dismissAlert()');
+    logger.debug('Alert banished!');
+    return true;
+  } else {
```
jlipps: following the same style, no need for the `else` here

lib/commands/context.js
```diff
@@ -0,0 +1,461 @@
...
+    return false;
+  }
+};
+
+helpers.stopRemote = async function (closeWindowBeforeDisconnecting = false) {
+  if (!this.remote) {
+    logger.errorAndThrow('Tried to leave a web frame but were not in one');
+  } else {
```
jlipps: same here

lib/commands/web.js
```diff
@@ -0,0 +1,201 @@
...
+
+commands.setUrl = async function (url) {
+  logger.debug(`Attempting to set url '${url}'`);
+  if (this.isWebContext()) {
+    // make sure to clear out any leftover web frames
+    this.curWebFrames = [];
+    this.processingRemoteCmd = true;
+    await this.remote.navToUrl(url);
+    this.processingRemoteCmd = false;
+  } else {
```
jlipps: same here and in all following. would prefer this to read:

```js
if (!this.isWebContext()) {
  throw new errors.NotImplementedError();
}
// rest of code goes here
```

lib/commands/web.js
```diff
@@ -0,0 +1,201 @@
...
+    return url;
+  } else {
+    throw new errors.NotImplementedError();
+  }
+};
+
+commands.title = async function () {
+  if (this.isWebContext()) {
+    this.processingRemoteCmd = true;
+    let title =  await this.executeAtom('title', []);
```
jlipps: doesn't executeAtom already set the `processingRemoteCmd`?

imurchie:
Alrighty! Thanks for the comments. It looks much better now.

lib/commands/context.js
```diff
@@ -0,0 +1,462 @@
...
+        logger.debug(`Remote debugger is already connected to window '${name}'`);
+      } else if (_.contains(_.pluck(this.windowHandleCache, 'id'), name)) {
+        await this.remote.disconnect();
+        this.curContext = this.curWindowHandle = name;
+        await this.remote.connect(name);
+      } else {
+        throw new errors.NoSuchWindowError();
+      }
+    }
+  } else {
```
jlipps: hate to nitpick, but this is popping out at me as needing to go up top as well :-) just seems natural to have your chain of "if this isn't the case, then throw an error" up top, with the meat below

lib/commands/context.js
```diff
@@ -0,0 +1,462 @@
...
+    if (this.isRealDevice()) {
+      pageArray = await this.remote.pageArrayFromJson();
+    } else {
+      pageArray = await this.remote.selectApp(this.caps.webviewConnectRetries);
+    }
+  } else {
+    if (this.isRealDevice()) {
+      this.remote = new WebKitRemoteDebugger();
+      return this.remote.pageArrayFromJson();
+    } else {
```
jlipps: because of the return above, don't need an else block here

lib/commands/context.js
```diff
@@ -0,0 +1,462 @@
...
+    }
+  };
+  await spinHandles();
+};
+
+helpers.closeAlertBeforeTest = async function () {
+  let present = await this.uiAutoClient.sendCommand('au.alertIsPresent()');
+  if (!present) {
+    return false;
+  } else {
```
jlipps: no need for an else with the return

lib/commands/web.js
```diff
@@ -0,0 +1,189 @@
...
+
+const ELEMENT_OFFSET = 5000;
+
+commands.setUrl = async function (url) {
+  logger.debug(`Attempting to set url '${url}'`);
+  if (!this.isWebContext()) {
+    // in the future, detect whether we have a UIWebView that we can use to
+    // make sense of this command. For now, and otherwise, it's a no-op
+    throw new errors.NotImplementedError();
+  } else {
```
jlipps: no need for else with throw

jlipps: and following (that's what i had in mind originally, and the benefit---one less level of indentation! one less pair of braces!)

lib/commands/context.js
imurchie: No need to hate nitpicking.

lib/commands/web.js
```diff
@@ -0,0 +1,189 @@
...
+    }
+    return element;
+  }
+};
+
+commands.click = async function (elementId) {
+  if (this.isWebContext()) {
+    let atomsElement = this.useAtomsElement(elementId);
+    return await this.executeAtom('click', [atomsElement]);
+  } else {
```
jlipps: basically just look anywhere you have an `else` in the whole project and chances are you can remove it :-)

PR #17, opened 2015-09-05
Safari implementation

imurchie:
There is cleanup to be done, but this is Safari for simulators.

Two parts don't work:
1 asynchronous execute of JS in a frame
2. performance logging

I will look into these under a separate heading.

### DIFF
--- a/lib/commands/gesture.js
+++ b/lib/commands/gesture.js
@@ -15,7 +15,8 @@ commands.nativeTap = async function (el) {
 commands.click = async function (el) {
   el = unwrapEl(el);
   if (this.isWebContext()) {
-    await this.webClick(el);
+    let atomsElement = this.useAtomsElement(el);
+    return await this.executeAtom('click', [atomsElement]);
   } else {
     if (this.opts.useRobot) {
       /* TODO */throw new errors.NotYetImplementedError();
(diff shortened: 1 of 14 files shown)


imurchie:
Fixed (1) above.

sebv:
What's the commented merge part for?

imurchie:
It was commented out, rather than deleted, while merging, since I didn't know where you had changed things. I need to delete it now.

lib/commands/web.js
```diff
@@ -53,39 +196,201 @@ commands.findWebElementOrElements = async function (strategy, selector, ctx, man
...
+  // add static offset for safari in landscape mode
+  let yOffset = this.curOrientation === 'LANDSCAPE' ? this.landscapeWebCoordsOffset : 0;
+
+  // absolutize web coords
+  let webviews = await this.uiAutoClient.sendCommand(wvCmd);
+  if (webviews.length < 1) {
+    throw new errors.UnknownError.code('Could not find any webviews to click inside!');
+  }
+  if (_.isUndefined(webviews[webviewIndex])) {
+    logger.warn("Could not find webview at index " + webviewIndex + ", " +
```
scottdixon: ${webviewIndex}

lib/commands/gesture.js
```diff
@@ -27,7 +28,10 @@ commands.click = async function (el) {
 
 commands.clickCurrent = async function (/*button*/) {
   if (this.isWebContext()) {
-    /* TODO */throw new errors.NotYetImplementedError();
+    if (_.isNull(this.curWebCoords)) {
```
jlipps: random comment. I think I prefer `this.curWebCoors === null`, unless `_.isNull` does something I'm not aware of

PR #19, opened 2015-09-16
Add support for safari on real device

imurchie:


lib/safari-launcher.js
```diff
@@ -0,0 +1,105 @@
...
+const SAFARI_LAUNCHER_BUNDLE = 'com.bytearc.SafariLauncher';
+
+const sdks = ['iphoneos'];
+
+async function cleanApp (appRoot, sdk) {
+  logger.debug(`Cleaning SafariLauncher for ${sdk}`);
+  try {
+    await exec('xcodebuild', ['-sdk', sdk, 'clean'], {cwd: appRoot});
+  } catch (err) {
+    logger.error(err);
```
jlipps: this looks like a good place for `logger.errorAndThrow` or whatever it is called

lib/safari-launcher.js
```diff
@@ -0,0 +1,105 @@
...
+
+async function buildApp(appRoot, sdk) {
+  try {
+    logger.debug(`Building SafariLauncher for ${sdk}`);
+    let args = ['-sdk', sdk, '-xcconfig', SAFARI_LAUNCHER_CONFIG_FILE];
+    await exec('xcodebuild', args, {
+      cwd: appRoot
+    });
+  } catch (err) {
+    logger.error(err);
```
jlipps: and here

PR #29, opened 2015-10-20
1.5 Cleanup

scottdixon:
TravisCI
Coverage
Readme
Completed todos:
- Deprecate iOS 7.x
- Removed endSimulatorDeamons, importing from ios-simulator
- Removed getDeviceString, importing from ios-simulator


lib/utils.js
```diff
@@ -1,14 +1,15 @@
...
 import _which from 'which';
 import B from 'bluebird';
 import { exec } from 'teen_process';
-import { waitForCondition } from 'asyncbox';
 import { SAFARI_BUNDLE } from './commands/safari';
 import { SAFARI_LAUNCHER_APP_FILE, SAFARI_LAUNCHER_BUNDLE } from './safari-launcher';
+import { endAllSimulatorDaemons } from 'appium-ios-simulator';
+
 
 const which = B.promisify(_which);
```
imurchie: `appium-support` now has `which` in `fs`.

lib/utils.js
```diff
@@ -384,5 +232,5 @@ function unwrapEl (el) {
 
 export default { rootDir, removeInstrumentsSocket, getAndCheckXcodeVersion, prepareIosOpts,
   appIsPackageOrBundle, getAndCheckIosSdkVersion,  detectUdid, parseLocalizableStrings,
-  setBundleIdFromApp, shouldPrelaunchSimulator, setDeviceTypeInInfoPlist, getDeviceString,
-  endSimulatorDaemons, getSimForDeviceString, unwrapEl };
+  setBundleIdFromApp, shouldPrelaunchSimulator, setDeviceTypeInInfoPlist, endAllSimulatorDaemons,
```
imurchie: Why is this exporting `endAllSimulatorDaemons`? Can't we just use it directly?

imurchie:
This is looking good. A couple of comments above.

Also: can we export a default `IosDriver `? In addition to the non-default, so we don't break anything?

imurchie:
This all looks reasonable to me.

PR #33, opened 2015-10-26
moved getBundleIdFromApp to its own module (so webdriverAgent-driver …

Jonahss:
…can use it too)

@imurchie

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,5 +1,6 @@
 import { fs, plist } from 'appium-support';
 import xcode from 'appium-xcode';
+import { extractBundleId } from 'ios-app-utils';
 import logger from './logger';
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -171,24 +172,11 @@ async function parseLocalizableStrings (opts) {
   }
 }
 
-async function getBundleIdFromApp (app) {
-  logger.debug("Getting bundle ID from app");
```
imurchie: Can you add a log line that would replace this. It would be good to have the bundleId in the log message, too, for posterity.

Jonahss:
@imurchie added a log line

imurchie:
:+1:

PR #34, opened 2015-10-29
TravisCI - parellel tests

scottdixon:
- Fixes TravisCI instruments issue.
- Runs TestApp and UICatalog tests side by side, takes ~30 mins total.

@imurchie

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 node_modules
 *.log
 test/assets
+*.plist
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,5 @@ before_script:
   - npm install -g mocha
 script:
   - gulp transpile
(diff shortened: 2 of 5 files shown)


scottdixon:
Testing Travis - please don't merge this.

jlipps:
scott let us know when this is ready for review. for example i wouldn't want the instruments files in the root dir

scottdixon:
I'd also like your thoughts on the best way to run coveralls at the end. There's no 'after_all_success' or similar option, hence the `FINAL_GULP=coveralls`. WWJLD?

jlipps:
that's an interesting strategy. i actually haven't set coveralls up on CI yet so a better person to ask would be @sebv

imurchie:
I'm going to merge this and then fix a few things.

PR #40, opened 2015-11-11
Port safari tests

christian-bromann:
ping @imurchie 

Moved Safari tests from
https://github.com/appium/appium/tree/master/test/functional/ios/safari
to
https://github.com/appium/appium-ios-driver/tree/port-safari-tests/test/e2e/safari

I had to quarantine the following tests because they were either not working or not supported anymore:
- _"touch should flick element"_ -> not working
- _"safari - webview - basic should check if two elements are equals"_ -> got different internal ids everytime
- _"safari - webview - basic should retrieve tag name of an element"_ -> haven't found a command to retrieve a tag name
- _"uicatalog - controls should be able to get and set a slider value"_ -> started failing for whatever reason

#### Possible improvements

The build is taking too long. I would suggest to run more tests concurrently but I am not sure what the best approach would be to achieve this. If we don't necessarily need to use the internal driver ([which gets overwritten anyway](https://github.com/appium/appium-ios-driver/blob/master/test/e2e/helpers/session.js#L16-L26)) we could use other tools like Webdriver.IO for these integration tests.


test/e2e/helpers/webview.js
```diff
@@ -0,0 +1,98 @@
...
+        throw new Error(`spinWait condition unfulfilled. Promise rejected with error: ${err}`);
+      }
+
+      return setTimeout(async () => await spin(spinFn), intMs);
+    }
+  };
+  await spin(spinFn);
+}
+
+function isChrome (desired) {
```
imurchie: This is unnecessary. We don't support Chrome on iOS.

test/e2e/helpers/webview.js
```diff
@@ -0,0 +1,98 @@
...
+let testEndpoint = function (desired) {
+  return isChrome(desired) ? env.CHROME_TEST_END_POINT : env.TEST_END_POINT;
+};
+
+function guineaEndpoint (desired) {
+  return isChrome(desired) ? env.CHROME_GUINEA_TEST_END_POINT :
+                             env.GUINEA_TEST_END_POINT;
+}
+
+// TODO: investigate why we need that
```
imurchie: Unnecessary (Android only).

imurchie:
Yes, that is taking a long time! And the webview tests are timing out. We can restrict some in the short term.

imurchie:
Closed. Used code in #44

PR #42, opened 2015-11-20
Better way for selecting address bar

christian-bromann:
Original PR: appium/appium#5836
Commit by @moizjv https://github.com/appium/appium/commit/c7610aecf58f47de3d19e620e73340edfe1f2bdd

Not sure about the other changes in that PR


imurchie:
We'll need to experiment to see if the other part is still necessary.

PR #46, opened 2015-11-30
Add plumbing for settings api, so tests don't fail

imurchie:
The base driver expects a `this.settings` property, or it fails. Even if the settings don't do anything.


jlipps:
:+1:

PR #52, opened 2015-12-09
Alert specs fix

scottdixon:
~~getAlertText should return alert text (not title), as per: http://www.w3.org/TR/webdriver/#dfn-get-alert-text~~


scottdixon:
Further research: getAlertText should return _both_ the title and subtext. Applying fix to `appium-uiauto`. https://github.com/appium/appium-uiauto/pull/95

PR #54, opened 2015-12-16
Add end test coverage method

imurchie:


jlipps:
isn't a NYI the default when a command isn't defined? i.e., isn't this essentially a no-op change?

PR #55, opened 2015-12-16
getSession: merge capabilities with server capabilities

scottdixon:


scottdixon:
@imurchie

imurchie:
LGTM

lib/driver.js
```diff
@@ -30,7 +30,7 @@ let iDevice = function (...args) {
 class IosDriver extends BaseDriver {
   resetIos () {
     this.appExt = ".app";
-    _.defaults(this.opts, {
+    this.serverDetails = {
```
jlipps: just curious, why do we create a field on the object for this instead of just using a local variable? i'm all for having as few member fields as possible. i.e., if it's not keeping any essential state, let's not remember it.

scottdixon: We need them again down on line 104/105.

We could remove it if this.opts doesn't actually care about server details?

scottdixon: Oh, I see - we could still avoid remembering them.

scottdixon:
@jlipps - variable tweaked!

jlipps:
LGTM

PR #59, opened 2016-01-05
added new keyword to get device date and time

SrinivasanTarget:
Hope this will fix https://github.com/appium/appium/issues/5893

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -2,6 +2,8 @@ import { errors } from 'mobile-json-wire-protocol';
 import _ from 'lodash';
 import js2xml from "js2xmlparser2";
 import log from '../logger';
+import { fs } from 'appium-support';
+import { exec } from 'teen_process';
 
 let commands = {}, helpers = {}, extensions = {};
 
@@ -10,6 +12,24 @@ commands.active = async function () {
     return await this.executeAtom('active_element', []);
   } else {
     return await this.uiAutoClient.sendCommand("au.getActiveElement()");
+  }
+};
+
+commands.getDeviceDateAndTime = async function() {
+  log.info('Attempting to capture iOS device date and time');
+  try {
+    if (this.isRealDevice()) {
+      let idevicedate = await fs.which('idevicedate');
+      log.info(`Found idevicedate: '${idevicedate}'`);
+      let {stdout, stderr, code} = await exec('idevicedate', ['-u', this.opts.udid]);
+      return stdout.trim();
+    } else {
+      log.warn('getDeviceDateAndTime command works only for real devices');
+       throw new errors.NotYetImplementedError();
+    }
+  } catch (err) {
+    log.warn('Could not capture device date and time using libimobiledevice idevicedate. ' +
+                'Libimobiledevice is probably not installed');
   }
 };
 


lib/commands/general.js
```diff
@@ -13,6 +15,27 @@ commands.active = async function () {
...
+      let idevicedate = await fs.which('idevicedate');
+      log.info(`Found idevicedate: '${idevicedate}'`);
+      let proc = new SubProcess('idevicedate', ['-u', this.udid], {env: spawnEnv});
+      proc.on('output', stdout => {
+        if (stdout) {
+        idevicedate = stdout.trim();
+        }
+      });
+      await proc.start();
+      await proc.stop();
```
Jonahss: I think for this case we can use teen_process.exec instead of SubProcess. `exec` is just like SubProcess but instead of manually starting and stopping, you can just await it once.

Jonahss:
Hey awesome, thanks for the contribution!
I left a small comment, with a suggestion of how to simplify the code. Could you update it?

After this gets merged, we will have to add the route to the server, to activate this method.

You can submit a separate pull request to this repo: https://github.com/appium/node-mobile-json-wire-protocol

I think we can make the route something like /device/system_time

Do you want to implement this on android driver too? Probably easy with adb. http://stackoverflow.com/questions/18135966/how-to-get-android-system-uptime-and-realtime-from-command-line

SrinivasanTarget:
Yes, ofcourse @Jonahss. Working on the changes suggested.Will update the PR in sometime. Will look into android and routes after this got merged.

Jonahss:
awesome!

SrinivasanTarget:
@Jonahss can you please review this?

lib/commands/general.js
```diff
@@ -13,6 +16,20 @@ commands.active = async function () {
   }
 };
 
+commands.getDeviceDateAndTime = async function(){
```
imurchie: Add space around function declaration's parentheses (i.e., `function () {`).

lib/commands/general.js
```diff
@@ -13,6 +16,20 @@ commands.active = async function () {
...
+    let udid = utils.detectUdid(auto);
+    let idevicedate = await fs.which('idevicedate');
+    log.info(`Found idevicedate: '${idevicedate}'`);
+    let {stdout, stderr, code} = await exec('idevicedate', ['-u', udid]);
+    return stdout.trim();
+  } catch (err) {
+    log.warn('Could not capture device date and time using libimobiledevice idevicedate. ' +
+                'Libimobiledevice is probably not installed');
+  }
+}
```
imurchie: Add semicolon.

imurchie:
Added a couple of style fixes. Otherwise, excellent!

Jonahss:
Getting closer, sorry for all the corrections.

Rather than using `util.detectUdid()`, we can use `this.opts.udid`. The driver object already is running a session when this command is called, so we already have a reference to the udid for this test.

We should also add a check, so make sure this command is only run on real devices (idevice_id) will not work with simulators I'm pretty sure. You can call `await this.isRealDevice()` and throw an error if it's false.

Jonahss:
oops, its just `this.isRealDevice()` without the `await`

SrinivasanTarget:
@Jonahss Is it fine now?

SrinivasanTarget:
@Jonahss Did get clear about the reason for CI job failure.

Jonahss:
Ah, I've seen this error elsewhere, I don't think it's your fault. will rerun.

SrinivasanTarget:
@Jonahss Build passed.

Jonahss:
Along with the log warning, can you also throw a NotImplemented error when realDevice is false? That way the test writer knows that the command failed, otherwise it will seem like it succeeds but the time returned will be null. This could result in an annoying null pointer exception for the end user.

SrinivasanTarget:
Yup Got it @Jonahss Included the changes

SrinivasanTarget:
@Jonahss Did'nt get clear about the reason for CI job failure.can you trigger again pls

SrinivasanTarget:
@imurchie Is this looks fine?

Jonahss:
I'm looking into it. This could either be an issue related to travis, or to appium-ios-simulator.
Currently working on figuring out whats wrong and trying to fix it.

SrinivasanTarget:
Any update on this @Jonahss

Jonahss:
I'm going to merge this!

Will be reviewing mjsonwp and android-driver related pull requests.

SrinivasanTarget:
Thanks @Jonahss Working on android-driver implementation.Will raise a PR for android-driver soon.

SrinivasanTarget:
@Jonahss Can you please review my android-driver and mjsonwp PR related to this

PR #61, opened 2016-01-06
Fix background for iOS 9+

imurchie:
Port of the bulk of the background fix from @wuhao5 (https://github.com/appium/appium/pull/5950)


wuhao5:
awesome! 

unrelated: is this to modularize the ios controller out of appium?

imurchie:
Yes. The new architecture moves the actual drivers out of the main package.

PR #65, opened 2016-01-12
Code refractoring and unit testing for getDeviceTime

SrinivasanTarget:
@Jonahss ping

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,79 +145,9 @@ console.log(await driver.getOrientation()); // -> 'LANDSCAPE'
 | `webContextIndex`                 |
 | `webFlickElement`                 |
 | `xySpeedFlick`                    |
-| `getDeviceTime`					| 
-| `getLatestWebviewContextForTitle` | 
-| `getLocation`                     | 
-| `getLocationInView`               | 
-| `getLog`                          | 
-| `getLogTypes`                     | 
-| `getName`                         | 
-| `getOrientation`                  | 
-| `getPageSource`                   | 
-| `getScreenshot`                   | 
-| `getSimFileFullPath`              | 
-| `getSize`                         | 
-| `getText`                         | 
-| `getUrl`                          | 
-| `getWindowHandle`                 | 
-| `getWindowHandles`                | 
-| `getWindowSize`                   | 
-| `handleTap`                       | 
-| `hideKeyboard`                    | 
-| `initAutoWebview`                 | 
-| `isWebContext`                    | 
-| `keys`                            | 
-| `launchApp`                       | 
-| `listWebFrames`                   | 
-| `lock`                            | 
-| `mobileRotation`                  | 
-| `mobileScroll`                    | 
-| `mobileShake`                     | 
-| `mobileWebNav`                    | 
-| `moveTo`                          | 
-| `nativeTap`                       | 
-| `nativeWebTap`                    | 
-| `navToInitialWebview`             | 
-| `navToViewThroughFavorites`       | 
-| `navToViewWithTitle`              | 
-| `onPageChange`                    | 
-| `parseElementResponse`            | 
-| `parseExecuteResponse`            | 
-| `parseTouch`                      | 
-| `performMultiAction`              | 
-| `performTouch`                    | 
-| `postAcceptAlert`                 | 
-| `postDismissAlert`                | 
-| `pullFile`                        | 
-| `pullFolder`                      | 
-| `pushFile`                        | 
-| `receiveAsyncResponse`            | 
-| `refresh`                         | 
-| `setAlertText`                    | 
-| `setContext`                      | 
-| `setCookie`                       | 
-| `setFrame`                        | 
-| `setGeoLocation`                  | 
-| `setOrientation`                  | 
-| `setUrl`                          | 
-| `setValue`                        | 
-| `setValueImmediate`               | 
-| `setWindow`                       | 
-| `stopRemote`                      | 
-| `submit`                          | 
-| `title`                           | 
-| `translateWebCoords`              | 
-| `typeAndNavToUrl`                 | 
-| `useAtomsElement`                 | 
-| `useNewSafari`                    | 
-| `waitForAtom`                     | 
-| `webContextIndex`                 | 
-| `webFlickElement`                 | 
-| `xySpeedFlick`                    | 
 
 
-## Watch
->>>>>>> updated command name as per jlips comments
+## Watch code for changes, re-transpile and run unit tests:
 
 ```
 gulp
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ console.log(await driver.getOrientation()); // -> 'LANDSCAPE'
 | `getCookies`                      |
 | `getCssProperty`                  |
 | `getCurrentContext`               |
-| `getDeviceDateAndTime`            |
+| `getDeviceTime`                   |
 | `getLatestWebviewContextForTitle` |
 | `getLocation`                     |
 | `getLocationInView`               |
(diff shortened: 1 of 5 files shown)


SrinivasanTarget:
https://github.com/appium/node-mobile-json-wire-protocol/pull/38

lib/commands/general.js
```diff
@@ -15,7 +15,7 @@ commands.active = async function () {
   }
 };
 
-commands.getDeviceDateAndTime = async function() {
+commands.getDeviceTime = async function() {
```
imurchie: There is one further change needed, in the log message on line 27.

README.md
```diff
@@ -76,7 +76,7 @@ console.log(await driver.getOrientation()); // -> 'LANDSCAPE'
 | `getCookies`                      | 
 | `getCssProperty`                  | 
 | `getCurrentContext`               |
-| `getDeviceDateAndTime`            | 
+| `getDeviceTime`       		    |
```
imurchie: Spacing is weird here.

imurchie:
Are there no tests for this feature?

README.md
SrinivasanTarget: Spacing looks fine in my sublime.Not sure what's going wrong here!. Will check and update again

SrinivasanTarget:
No @imurchie Will definitely add test in sometime. Checking out how to add test.

imurchie:
Rerunning the failed tests (which probably don't have anything to do with this change).

SrinivasanTarget:
@imurchie Added tests please review

imurchie:
Another commit broke our CI, so tests are going to fail for now.

But:

As an e2e test this will only work on a real device. It would be better to have a unit test, so that we can stub the calls and make sure that the method functions correctly.

Finally, after having run this locally, it becomes clear that the function needs to be changed a bit. As it stands, the exception thrown by the simulator case is caught immediately, so we end up with `undefined` as the return value. The try/catch block should only be the real device case. And perhaps an exception should be thrown if it fails.

```js
commands.getDeviceTime = async function () {
  log.info('Attempting to capture iOS device date and time');
  if (this.isRealDevice()) {
    try {
      let idevicedate = await fs.which('idevicedate');
      log.info(`Found idevicedate: '${idevicedate}'`);
      let {stdout} = await exec('idevicedate', ['-u', this.opts.udid]);
      return stdout.trim();
    } catch (err) {
      log.errorAndThrow('Could not capture device date and time using libimobiledevice idevicedate. ' +
                     'Libimobiledevice is probably not installed');
    }
  } else {
    log.warn('getDeviceDateAndTime command works only for real devices');
    throw new errors.NotYetImplementedError();
  }
};
```

SrinivasanTarget:
@imurchie Thanks for the detailed comment.I got the changes now. Can you please help me in writing unit test with any example available?

SrinivasanTarget:
@imurchie Can you please review this?

imurchie:
If this is a unit test, let's move it over into the unit tests (https://github.com/appium/appium-ios-driver/blob/master/test/driver-specs.js).

It then does not need to create a session or anything. Just instantiate the driver object, and call the `getDeviceTime` method.

We need to test two conditions: a real device and a simulator. In order to do this, `IosDriver#isRealDevice` needs to be stubbed in order to return `true` or `false` when we want it. For the simulator case, this is enough, and we would expect to get a `NotYetImplementedError` exception thrown.

For the real device case, when our stub of `IosDriver#isRealDevice` returns `true`, we need a little more. First we need to stub `fs.which` so that we don't need to actually call out to the file system. It doesn't matter what that returns (except that it should be a Promise), because we also need to stub out `exec` so we are not actually running the `idevicedate` program.

The `exec` stub can, for one test, return a date string, which we will expect, and in another test it can throw an exception, which we catch and pass along.

An example of mocking (which is like stubbing, and we have some utilities for it) `fs`. It can be adapted for `fs.which()`: https://github.com/appium/appium-adb/blob/master/test/unit/helper-specs.js#L30

An example of mocking `teen_process#exec`: https://github.com/appium/appium-adb/blob/master/test/unit/android-manifest-specs.js#L38

imurchie:
A stab at the successful real device case (showing the whole driver spec file):

```js
// transpile:mocha

import IosDriver from '../';
import chai from 'chai';
import chaiAsPromised from 'chai-as-promised';
import { withMocks } from 'appium-test-support';
import * as teen_process from 'teen_process';
import { fs } from 'appium-support';


chai.should();
chai.use(chaiAsPromised);

describe('driver', () => {
  it('should instantiate class', () => {
    let driver = new IosDriver();
    driver.should.exist;
  });

  describe('getDeviceTime', withMocks({fs, teen_process}, (mocks) => {
    it('should return a valid date string', async () => {
      let udid = 'some-udid';
      let date = new Date().toString();
      mocks.fs.expects('which')
        .once().returns('/path/to/idevicedate');
      mocks.teen_process.expects('exec')
        .once().withExactArgs('idevicedate', ['-u', udid])
        .returns({stdout: date});
      let driver = new IosDriver();
      driver.opts = {udid};

      (await driver.getDeviceTime()).should.equal(date);
    });
  }));
});
```

SrinivasanTarget:
@imurchie Thanks for detailed explanation. It's a great learning for me!
Have included unit testing now.

test/driver-specs.js
```diff
@@ -14,3 +17,35 @@ describe('driver', () => {
...
+      let driver = new IosDriver();
+      driver.opts = {udid};
+
+      (await driver.getDeviceTime()).should.equal(date);
+    });
+    it('should not return a valid date string', async () => {
+      let udid = 'some-udid';
+      let date = new Date().toString();
+      mocks.fs.expects('which')
+        .once().returns('/path/to/idevicedate');
```
imurchie: This test ought to do the case of no `libimobiledevice` installed. That would mean that `exec` would throw an error saying that it couldn't find `idevicedate`.

You can mock this like:

```js
mocks.teen_process.expects("exec")
  .once()
  .throws("ENOENT");
let driver = new IosDriver();
driver.opts = {udid};
await adb.getDeviceTime()
  .should.eventually.be.rejectedWith("Could not capture device date and time");
```

imurchie:
Can you run, on your machine, run `gulp jshint`, and fix the problems that come up?

Then try out the tests you are running by running `gulp once` (or, while you are editing, you can have `gulp watch` running, which will rerun the unit tests each time you save your document).

SrinivasanTarget:
@imurchie please review

imurchie:
CI is failing because you need to add `appium-test-support` to the dependencies. You can run `npm install --save-dev appium-test-support` and then add `package.json` to your commit, and push.

This will allow the tests to use the `withMocks` function.

SrinivasanTarget:
@imurchie Im not able to squash my commits and unfortunately it includes a merge commit as well.Rebase is not helping me.Please help

imurchie:
What happens in the rebase?

SrinivasanTarget:
Interactive rebase hungs while squashing

imurchie:
Is there any output?

SrinivasanTarget:
Nope two or three commits gets squashed and then it leaves conflicts in branch message

SrinivasanTarget:
Now Tests got passed, does test code look fine now?

imurchie:
Yes, the code looks good now. I'm not sure why rebase is failing, though I think it is probably something circular happening in the merge commit.

SrinivasanTarget:
It happened even before the merge commit:(

imurchie:
Hmm. At this point I would normally save the code, rollback, and put it in anew. :(

SrinivasanTarget:
Ok,will do that now.

SrinivasanTarget:
Closing this will raise a new PR:(

PR #67, opened 2016-01-12
Fix travis config after tests shifted

imurchie:
After the logging tests were moved travis began to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: required
 env:
   # Run tests in parellel, up to 5 at once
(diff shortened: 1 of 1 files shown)


jlipps:
this conflicts with what I just pushed; does your code here _only_ run unit tests or will it not run everything under test/?

imurchie:
It just runs the unit tests. We end up running `mocha build/test/ -g @skip-ci -i`

https://travis-ci.org/appium/appium-ios-driver/jobs/101911202

imurchie:
In your configuration we never run the unit tests at all.

imurchie:
In lieu of #68, closing.

PR #72, opened 2016-01-15
executeAsync fix + CI

scottdixon:
- Fixed `executeAsync` race condition (sometimes driver attempts to resolve promise before it's created).
- Including `executeAsync` in CI


scottdixon:
cc @imurchie

imurchie:
Awesome catch!

PR #73, opened 2016-01-19
Fix handling of servers

imurchie:
We were using a single server, attached to a session. This made it so subsequent sessions could not use the server, since the `IosDriver` object was wrong. This manifest itself in 404 errors while trying to do async executes.

This starts a new server with each session. Once this is in and the build is good, we can look into handling the server in a different way (most of our calls to the server are context-less, involving only getting static pages. It is only the execute async ones that call back to the object. Perhaps we can have a single generic server that can be replaced if necessary).

@scottdixon this ought to fix the `executeAsync` woes.


scottdixon:
:+1:

imurchie:
Updated to only start it new if we mark that as necessary.

imurchie:
Build was clean. Pushed version with a commented out section removed, so skipped build.

PR #74, opened 2016-01-19
code refractoring and unit testing

SrinivasanTarget:
Raising this instead of #65

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -16,21 +16,21 @@ commands.active = async function () {
   }
 };
 
-commands.getDeviceDateAndTime = async function() {
+commands.getDeviceTime = async function () {
   log.info('Attempting to capture iOS device date and time');
-  try {
-    if (this.isRealDevice()) {
+  if (this.isRealDevice()) {
+    try {
       let idevicedate = await fs.which('idevicedate');
       log.info(`Found idevicedate: '${idevicedate}'`);
       let {stdout} = await exec('idevicedate', ['-u', this.opts.udid]);
       return stdout.trim();
-    } else {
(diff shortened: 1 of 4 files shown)


SrinivasanTarget:
@imurchie FYI

lib/commands/general.js
```diff
@@ -67,7 +68,22 @@ commands.getPageSource = async function () {
 };
 
 commands.background = async function (secs) {
-  await this.uiAutoClient.sendCommand(`au.background(${secs})`);
```
imurchie: This is overwriting a change that is needed.

SrinivasanTarget:
Updated!

imurchie:
That function ought to look like https://github.com/appium/appium-ios-driver/blob/master/lib/commands/general.js#L69-L71

SrinivasanTarget:
Yes updated it now

SrinivasanTarget:
Squashed now

imurchie:
Looks good. We'll let CI run and then get it in! Sorry for the slowness of the process.

imurchie:
Awesome! Merging! Thank you for bearing with us through this.

SrinivasanTarget:
Thanks @imurchie What's the next step on this?Shall i proceed with java client implementation?

imurchie:
Yes. I will make an issue in each of the clients, to create a method to use this endpoint. You can begin the Java implementation if that is your language of choice.

SrinivasanTarget:
(y)

imurchie:
I've created an issue in the `mobile-json-wire-protocol` repo to track this progress.

https://github.com/appium/node-mobile-json-wire-protocol/issues/40

SrinivasanTarget:
Thanks @imurchie

PR #75, opened 2016-01-19
Don't skip safari tests in CI

imurchie:
A bunch of Safari tests were skipped because the tests were timing out. That doesn't happen anymore. So don't do it.


scottdixon:
`should execute async javascript in frame` suffers from the same 404 issue. Your setup reset fix resolves it. Tested locally with success.

imurchie:
Yup. Pushing that through.

imurchie:
Thanks, coveralls. 0.0% decrease.

PR #77, opened 2016-01-21
Restart simulator after we modify plist files

scottdixon:
`locationServicesAuthorize` was failing iOS tests. Forcing a restart after we modify plists resolved the issue.

It would always work the first time because plist files are modified before boot. Subsequent modifications wouldn't take effect.

Just ran all e2e-tests with success.

@imurchie - is this a sensible place for the restart?

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -586,6 +586,9 @@ class IosDriver extends BaseDriver {
       logger.error("Error setting safari preferences, prefs will not work");
       logger.error(e);
     }
+
+    logger.debug("Updated plist files, rebooting the simulator if it's already open");
+    await this.endSimulator();
   }
 
   async setLocServicesPrefs () {


imurchie:
I'm a little concerned that this is heavy-handed. If the sim is new, we launch and quit, then modify some plists, then launch and quit again. Then we launch for the actual test. If the sim is not new, we modify the plists, launch and quit, then we launch for the actual test.

Is the problem that the preferences take a launch and quit in order to take effect? Or that we are modifying them while the sim is running, and so they don't take effect?

If it is the latter, can we not just quit the simulator before we do the plist tweaking? Then it will be relaunched when we start the test proper?

Something like:

```js
if (await this.sim.isFresh()) {
  await this.launchAndQuitSimulator();
} else {
  await this.endSimulator();
}
```

scottdixon:
Thanks @imurchie - I think we can use `endSimulator()` for both cases - see update. e2e suite is happy with this.

Also, e2e tests were already picking up this failure. So I think we can ignore the coveralls complaint and add a bunch of unit tests as another ticket.

imurchie:
Cool!

And yes, I agree with the coverage issue. Plus, 0.02% decrease is minimal.

PR #91, opened 2016-02-03
Fix IWD bug

scottdixon:


jlipps:
@scottdixon :scream: 

the only other thing i'd like to see is a unit test that exercises `prepareIosOpts` with different values of nativeInstrumentsLib and asserts the right things.

moizjv:
:+1:

scottdixon:
Good call @jlipps - added tests.

PR #92, opened 2016-02-04
return correct information on create session

scottdixon:
Fixes https://github.com/appium/appium/issues/5920

createSession wasn't returning the `javascriptEnabled` server capability - ruby client wouldn't execute JS.

CC @imurchie @KazuCocoa


imurchie:
Interesting. Looks good to me.

KazuCocoa:
thanks! :+1:

PR #96, opened 2016-02-12
Add support for getDeviceTime on sims

imurchie:
Instead of throwing an error, send the date and time of the device running the sim.

cc @SrinivasanTarget


SrinivasanTarget:
(y)

PR #100, opened 2016-02-23
Change browserName for server capabilities

imurchie:
Rather than having an invalid `browserName` as a server capability, use `''`.

Addresses https://github.com/appium/appium/issues/6177


vania-pooh:
@imurchie could you give a short explanation why you consider browserName: 'iOS' as invalid?

PR #102, opened 2016-02-24
Fix handling of arguments to execute

imurchie:
We were only converting element arguments and skipping the rest. Fix that.


jlipps:
:+1:

PR #110, opened 2016-03-08
Log an error and continue if instruments doesn't shut down

scottdixon:
Supersedes https://github.com/appium/appium-instruments/pull/61

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -141,10 +141,19 @@ class IosDriver extends BaseDriver {
 
   async stop () {
     this.ready = false;
-    let stopPromises = [];
-    if (this.uiAutoClient) stopPromises.push(this.uiAutoClient.shutdown());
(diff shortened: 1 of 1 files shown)


lib/driver.js
```diff
@@ -141,10 +141,15 @@ class IosDriver extends BaseDriver {
 
   async stop () {
     this.ready = false;
-    let stopPromises = [];
-    if (this.uiAutoClient) stopPromises.push(this.uiAutoClient.shutdown());
-    if (this.instruments) stopPromises.push(this.instruments.shutdown());
-    await B.all(stopPromises);
+
+    if (this.uiAutoClient) await this.uiAutoClient.shutdown();
```
imurchie: Can we get this on its own line, in a block?

imurchie:
LGTM.

PR #116, opened 2016-03-23
Fix compatibility with appium-instruments 3.6.4

argon:
This fix removes parameters to a call which previously didn't take parameters but now does. The parameters passed by the driver cause the method call to fail.

Recent versions of `appium-instruments` have taken no parameters to `getAvailableDevices` so passing the `xcodeVersion` etc had no effect. With the latest version [`3.6.4`](https://github.com/appium/appium-instruments/compare/2fb29998df161fcb6f19ddbf27c9844acb747094...23d689b) the first parameter is a timeout value.

This results in the driver not working as it passes a `timeout` value of `7.2.1` on Xcode 7.2.1 causing an error.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -877,8 +877,7 @@ class IosDriver extends BaseDriver {
     }
 
     logger.debug("Checking whether instruments supports our device string");
-    let availDevices = await retry(3, instrumentsUtils.getAvailableDevices, this.xcodeVersion,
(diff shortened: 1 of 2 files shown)


imurchie:
This was fixed in #115 (published in `appium-ios-driver` 1.9.1). You caught it in between merges, unfortunately.

PR #118, opened 2016-04-01
fix set in opts ipa file

HlebHalkouski:
fix incorrect substring


imurchie:
Good catch. I messed that up copy and pasting from the old version. Thanks!

PR #124, opened 2016-04-12
Make sure getWindowHandles returns strings

imurchie:
The spec says it needs to be an array of strings. 

Resolves https://github.com/appium/appium/issues/6396


DylanLacey:
:+1:

PR #128, opened 2016-04-19
Make sure xpath respects implicit wait

imurchie:
The sub method should not throw an error.

Fixes https://github.com/appium/appium/issues/6411


imurchie:
Travis is broken. All tests pass locally. Merging.

PR #129, opened 2016-04-19
Update logging for reset

imurchie:
Just clean up logging a little so it is more clear what is happening.


imurchie:
Travis is broken. Tests pass locally, and this is a trivial change. Merging.

PR #131, opened 2016-04-26
add in-built app: Calendar's support.

purplum:
add in-built app support for Calendar.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -218,6 +218,12 @@ class IosDriver extends BaseDriver {
           this.opts.bundleId = "com.apple.Preferences";
           this.opts.app = null;
         }
+      } else if (this.opts.app && this.opts.app.toLowerCase() === "calendar") {
+        if (parseFloat(this.opts.platformVersion) >= 8) {
+          logger.debug("We are on iOS8+ so not copying calendar app");
+          this.opts.bundleId = "com.apple.mobilecal";
+          this.opts.app = null;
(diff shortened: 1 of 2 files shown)


imurchie:
Awesome! Thanks for the work in making this happen!

A simple test would be great to have. I've written one that would suffice, if you would like to add it to your PR. The file should be, I think, `test/e2e/driver/built-in-apps-e2e-specs.js` and the contents are here: https://gist.github.com/imurchie/6efbd0bd5a5e228d3b1e2b3dbf7ea8e3

purplum:
@imurchie , just added the e2etest case fetched from the repo you provided.
Thanks!

imurchie:
Whoops! I should have linted that first.

Can you run `gulp jshint` and fix the errors that come up, then re-push?

purplum:
checked the build log, seems it's using ios8.4, so need to change the deviceName from "iPhone 6s" to "iPhone 6".

purplum:
@imurchie Fixed now.

imurchie:
Thanks! Sorry for the test not being completely right.

One final thing: can you squash this into a single commit? Then I can merge.

purplum:
Opened a new PR with squash commit..
Close this one.

imurchie:
Ah! Sorry. For future reference, after you squash you can just force-push to the same branch and it will update here. But I'll track the other PR.

PR #132, opened 2016-04-27
Open new window if Safari URL bar is not found

imurchie:
When Safari is in the state where all the pages are visible, we cannot tap the URL bar, and so we cannot initialize a Safari session. In this case, find the "New page" button and tap on it to get into a useable Safari state.


jlipps:
👍

PR #134, opened 2016-04-28
Add logging if safari test is started with no keyboard

imurchie:
In the case that the user does not have the keyboard toggled on, Safari will not be able to load. Add logging for this case.


jlipps:
change LGTM when you've determined it's effective!

PR #136, opened 2016-04-28
Make sure keys gets arguments right

imurchie:
We get an array sometimes. Make sure we send a string along.

See https://github.com/appium/appium/issues/6441

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -116,6 +116,12 @@ commands.keys = async function (keys) {
     }
     await this.setValue(keys, el.ELEMENT);
   } else {
+    if (_.isArray(keys)) {
+      keys = keys.join('');
+    }
+    if (!_.isString(keys)) {
+      keys = `${keys}`;
+    }
     keys = util.escapeSpecialChars(keys, "'");
     let command = `au.sendKeysToActiveElement('${keys}')`;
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -116,6 +116,12 @@ commands.keys = async function (keys) {
     }
     await this.setValue(keys, el.ELEMENT);
   } else {
+    if (_.isArray(keys)) {
+      keys = keys.join('');
+    }
+    if (!_.isString(keys)) {
+      keys = `${keys}`;
```
jlipps: is this the same as `keys = keys.toString()`?

imurchie: It is, as long as `keys` is an array with a string inside. I think the spec says it ought to take an array of characters, in which case `['a', 'b', 'c'].toString() == 'a,b,c'`

jlipps: Hmm, I was more commenting on what we would prefer to standardize on if

    `${keys}`

and

    keys.toString()

are equivalent.

imurchie: Ah! I see. Yes, `toString` is better. And not sure why I went with the other, to be perfectly honest.

imurchie: https://github.com/appium/appium-ios-driver/pull/137

PR #139, opened 2016-05-02
Move the ipa check to after the archive has been unzipped

### LINKED ISSUE #138: Downloading an ipa from a url fails to install on a real device
When downloading an ipa from a url, the install fails when using a real device with the following error:
`[INST STDERR] Instruments Usage Error: Specified target process is invalid: [redacted]`

This is caused by the fix at https://github.com/appium/appium-ios-driver/blob/master/lib/driver.js#L209. This helped .ipa files to be installed correctly, but the BaseDriver unzips the .ipa and `opts.app` is now pointing at an .app file instead. This causes a mismatch of file paths between `opts.ipa` and `opts.app` and the ipa url is attempted to be installed (which doesn't fail).

```
[debug] [iOS] App is an ipa
[BaseDriver] Using downloadable app 'http://[redacted]/real_device_tests.ipa'
[debug] [BaseDriver] http://[redacted]/real_device_tests.ipa downloaded to /var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/appium-app.zip
[BaseDriver] Downloaded app to '/var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/appium-app.zip'
[debug] [BaseDriver] Unzipping /var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/appium-app.zip
[debug] [BaseDriver] Testing zip archive: /var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/appium-app.zip
[BaseDriver] Unzipped local app to '/var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/Payload/[redacted].app'
[debug] [iOS] Removing any remaining instruments sockets
...
...
...
[debug] [iOS] App is not installed. Will try to install the app.
[debug] [iOS] app: /var/folders/n_/d_bw41_d1fjd_nn22sdw_4d80000gn/T/201642-5150-148yoxr.65h9cc8fr/Payload/[redacted].app
[debug] [iOS] ipa: http://[redacted]/real_device_tests.ipa
[debug] [iOS] Installing ipa found at http://[redacted]/real_device_tests.ipa
[debug] [iOS] Nothing found on device, going ahead and installing.
```
Note: [debug] [iOS] App is an ipa, app: [path], and ipa: [path] were logs added by me for testing

JMoravec:
fixes #138 

I'm not sure if this is the best solution, but testing this locally worked for me.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -204,14 +204,6 @@ class IosDriver extends BaseDriver {
         this.opts.bundleId = this.opts.app;
       }
 
-      // if we have an ipa file, set it in opts
-      if (this.opts.app) {
-        let ext = this.opts.app.substring(this.opts.app.length - 3).toLowerCase();
-        if (ext === 'ipa') {
-          this.opts.ipa = this.opts.app;
-        }
-      }
-
       if (this.opts.app && this.opts.app.toLowerCase() === "settings") {
         if (parseFloat(this.opts.platformVersion) >= 8) {
           logger.debug("We are on iOS8+ so not copying preferences app");
@@ -331,6 +323,14 @@ class IosDriver extends BaseDriver {
     if (this.opts.autoLaunch === false) {
       return;
     }
+
+      // if we have an ipa file, set it in opts
+      if (this.opts.app) {
+        let ext = this.opts.app.substring(this.opts.app.length - 3).toLowerCase();
+        if (ext === 'ipa') {
+          this.opts.ipa = this.opts.app;
+        }
+      }
 
     if (this.opts.udid) {
       if (await this.realDevice.isInstalled(this.opts.bundleId)) {


lib/driver.js
```diff
@@ -332,6 +324,14 @@ class IosDriver extends BaseDriver {
       return;
     }
 
+      // if we have an ipa file, set it in opts
+      if (this.opts.app) {
```
jlipps: I almost want to say that if we are in this method and _don't_ have an ipa at this point, that is a problem, no? Given that we can only install ipas to real devices?

I wonder if the whole distinction between `this.opts.app` and `this.opts.ipa` even makes sense anymore?

jlipps: oh. I guess we can install .apps as well as .ipas.

Still, something seems off architecturally here. I'm fine with your fix but I think it highlights a pre-existing bit of conceptual debt. @imurchie what do you think?

imurchie: Yes, this looks good to me. There is no reason to do the ipa check earlier. 

As for conceptual debt, I agree. This was largely copied over from the old code. It might be good to encapsulate this logic into an `App` class or something. But it is more of a change than a bugfix, and given our current structure, this patch is good.

JMoravec: Looking into this a little deeper, is there a reason there is a seperate `installIpa` method? I could be missing something, but it lloks like .ipa's are always unziped in the BaseDriver: 
https://github.com/appium/appium-base-driver/blob/b1323602d46fc7967913758e3b2a913d1b0f9bc8/lib/helpers.js#L22

imurchie: Hmm. `installIpa` also bypasses any reset/no reset logic we have, and always uninstalls. Weird. I think we need to look at the install flow altogether.

imurchie:
Restarting the Travis job. For some reason the ios-test-app didn't install.

imurchie:
Merging this. Added discussion of install flow to https://github.com/appium/appium-ios-driver/issues/141

PR #140, opened 2016-05-03
Fix simulator error when language is set

rgonalo:
Fix https://github.com/appium/appium/issues/6435


imurchie:
That is precisely the sort of error it seemed like. I couldn't find it myself. Thank you so much!

jlipps:
wow! we need some unresolved promise warnings :-) thanks @rgonalo !

PR #143, opened 2016-05-10
Consolidate safari tests to minimize sessions

imurchie:
We have a number of related tests that are in different sessions. Consolidate in the hopes of speeding up the Safari build.


.travis.yml
```diff
@@ -10,12 +10,6 @@ language:
   - objective-c
 osx_image: xcode6.4
 before_install:
-  # TavisCI's OSX environment gives us node 0.10.32
```
jlipps: travis gives us 0.12 by default now?

jlipps:
the diff is a bit hard to read but i like the idea and i'll trust you that all the tests are there!

PR #144, opened 2016-05-11
Allow no webkit proxy on connect

imurchie:
On real device tests, if you call `getContexts` and don't ever expect there to be one (in a native app, say), and therefore are not running the ios-webkit-debug-proxy, it will error out with a connection refused error. This is not what we want.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -173,8 +173,20 @@ extensions.listWebFrames = async function (useUrl = true) {
     pageArray = await this.remote.selectApp(currentUrl, this.opts.webviewConnectRetries);
   } else if (this.isRealDevice()) {
     // real device, and not connected
-    this.remote = new WebKitRemoteDebugger({port: this.opts.webkitDebugProxyPort});
-    pageArray = await this.remote.pageArrayFromJson();
+    try {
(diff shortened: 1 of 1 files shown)


lib/commands/context.js
```diff
@@ -173,8 +173,18 @@ extensions.listWebFrames = async function (useUrl = true) {
...
-    pageArray = await this.remote.pageArrayFromJson();
+    try {
+      this.remote = new WebKitRemoteDebugger({port: this.opts.webkitDebugProxyPort});
+      pageArray = await this.remote.pageArrayFromJson();
+    } catch (err) {
+      // it is reasonable to expect that this might be called when there is no
+      // webkit remote debugger to connect to
+      if (!_.includes(err.message, 'connect ECONNREFUSED')) throw err;
+
+      logger.debug('Unable to connect to ios-webkit-debug-proxy. Continuing as if that is correct');
```
jlipps: Wonder if we want to call this out in the logs more? E.g.:

```js
logger.warn('Attempted to get a list of webview contexts but could not connect to ' +
            'ios-webkit-debug-proxy. If you expect to find webviews, please ensure ' +
            'that the proxy is running and accessible');
```

imurchie: Sounds reasonable. :)

PR #147, opened 2016-05-16
Make sure scroll doesn't throw an error when it hits the end of a view

imurchie:
Otherwise oyu need to wrap every `mobile: scroll` call to catch an error when it hits the end of the view.


jlipps:
LGTM

PR #148, opened 2016-05-17
prevent getUrl from breaking getWindowHandles - resolves #6506

scottdixon:
Resolves https://github.com/appium/appium/issues/6506

When remote debugger receives a `PAGE_CHANGE` event, the `pageArray` is updated with a list of current URLs. This event is triggered if we navigate to a URL or reload. This event _isn't_ triggered if a fragment identifier is appended to the URL. E.g. http://google.com#blah

If we call `getUrl` after a fragment identifier is appended to the URL, the page array and the current url get out of sync, which [hinders `selectApp`](https://github.com/appium/appium-remote-debugger/blob/master/lib/remote-debugger.js#L183-L186).

All `getUrl` does is return the output of `window.location.href`

The simplest solution is to stop updating `currentUrl` when we `getUrl`.

Please review @imurchie

### DIFF
--- a/lib/commands/web.js
+++ b/lib/commands/web.js
@@ -78,7 +78,6 @@ commands.getUrl = async function () {
     throw new errors.NotImplementedError();
   }
   let url = await this.remote.execute('window.location.href');
-  this.setCurrentUrl(url);
   return url;
 };
 
--- a/test/e2e/safari/webview/basic-specs.js
+++ b/test/e2e/safari/webview/basic-specs.js
@@ -117,6 +117,13 @@ describe('safari - webview - basics', function() {
       (await driver.getUrl()).should.include('test/guinea-pig');
     });
 
+    it('should get updated URL without breaking window handles', async () => {
+      let el = await driver.findElement('link text', 'i am an anchor link');
+      await driver.click(el);
+      (await driver.getUrl()).should.contain('#anchor');
+      (await driver.getWindowHandles()).should.be.ok;
+    });
+
     it('should send keystrokes to specific element', async () => {
       let el = await driver.findElement('id', 'comments');
       await driver.clear(el);


jlipps:
@scottdixon why did we ever have `setCurrentUrl`? Was it useful for non-anchor changes? I wonder if we don't want to delete it but put it behind a conditional that checks to see if there's a `#` in the url? maybe @imurchie knows what the purpose was

scottdixon:
@jlipps I can only find reference to `_currentUrl` in `listWebFrames()`, which is where the bug stems from. `setCurrentUrl()` feels like a left-over from `setUrl()` (where it makes sense).

If `setCurrentUrl()` is _actually_ serving a purpose in the `getUrl()` method, placing it behind a conditional `if #` would still be bug prone.

The safest method is to update both the `_currentUrl` _and_ the `pageArray` on a `getUrl()` but I don't think all this logic is required for a simple `window.location.href` - we already have the URL details from the `PAGE_CHANGE`.

I wonder if HTML5's history API triggers a `PAGE_CHANGE`. Let me see.

I look forward to your thoughts, @imurchie

imurchie:
The current url was added because in iOS 9 we stopped being able to rely on the order of apps/pages coming in. We would _almost always_ end up choosing the wrong app, and failing. So it was a hack to allow us to select the app we wanted.

scottdixon:
@imurchie makes sense and I see it's place in setUrl/page_change. Is it useful in `getUrl()`? Or has this been added by mistake?

imurchie:
Not entirely sure.

jlipps:
OK. I just know this stuff is fiddly and I don't want to break things by removing something we don't understand. Guess we should understand why it's there more clearly?

scottdixon:
Further research - history.pushState() doesn't trigger a `PAGE_CHANGE`. I still think the current PR makes sense. I'll whiteboard my reasoning with you @imurchie to make sure I'm not missing anything.

jlipps:
sounds good, i'm all for less magic things we don't know work!

imurchie:
I'm happy with this. Merging.

PR #151, opened 2016-06-01
mkdirp not fs.mkdirp

imurchie:
Users have reported errors involving this (i.e., "undefined is not a function" errors).


jlipps:
👍  wow can we get this into 1.5.3?

imurchie:
I hope so, yes!

PR #154, opened 2016-06-23
Add error message when attempting to use Xcode 8

imurchie:
Don't allow requests to get too far, since they will not work.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -109,6 +109,16 @@ class IosDriver extends BaseDriver {
   async createSession (caps) {
     let [sessionId] = await super.createSession(caps);
 
+    // appium-ios-driver uses Instruments to automate the device
+    // but Xcode 8 does not have Instruments, so short circuit
+    this.xcodeVersion = await utils.getAndCheckXcodeVersion(this.opts);
+    logger.debug(`Xcode version set to ${this.xcodeVersion.versionString}`);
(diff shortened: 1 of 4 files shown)


lib/driver.js
```diff
@@ -109,6 +109,14 @@ class IosDriver extends BaseDriver {
   async createSession (caps) {
     let [sessionId] = await super.createSession(caps);
 
+    // appium-ios-driver uses Instruments to automate the device
+    // but Xcode 8 does not have Instruments, so short circuit
+    this.xcodeVersion = await utils.getAndCheckXcodeVersion(this.opts);
+    logger.debug(`Xcode version set to ${this.xcodeVersion.versionString}`);
+    if (this.xcodeVersion.major === 8) {
```
jlipps: to be future proof, what about a `>= 8`?

lib/driver.js
```diff
@@ -109,6 +109,14 @@ class IosDriver extends BaseDriver {
...
+      throw new errors.SessionNotCreatedError(`appium-ios-driver does not support xcode version ${this.xcodeVersion.versionString}.`);
```
jlipps: maybe a note as to _why_ it's not supported? (`Apple has deprecated UIAutomation. Use the "XCUITest" automationName capability instead`)

PR #155, opened 2016-06-23
Fix function spacing

imurchie:
Add a space before opening parenthesis in function declarations. Once this is in I'll update the eslint config to add the rule to check this.


jlipps:
👍

PR #156, opened 2016-06-28
fix the calling of deleted ncp function

bigfei:
fs.ncp is not exporting anymore. Use copyFile instead.


imurchie:
Thank you! This is a good catch!

PR #158, opened 2016-06-29
fix the curOrientation in various places.

bigfei:
According to the orientation.js, it should be checking the 
`this.opts.curOrientation` rather than the old `this.curOrientation`.


imurchie:
This seems reasonable to me. Thanks!

imurchie:
Thanks for this work!

PR #159, opened 2016-07-03
Update for newer form of base driver timeouts

imurchie:


package.json
```diff
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "adm-zip": "^0.4.7",
-    "appium-base-driver": "^2.0.3",
+    "appium-base-driver": "^2.0.9",
```
jlipps: was this a breaking change in base-driver?

jlipps:
LGTM

package.json
imurchie: It's not a breaking change. It's just something additional that can be made use of here.

jlipps: ok 👍

PR #162, opened 2016-07-19
Port of appium 1.4 swipe action

### LINKED ISSUE #150: Swipe works incorrect. WaitAction is not perform
Fast swipe, different behavior with Android.

## Environment
* Appium version (or git revision) that exhibits the issue: 1.5.2
* Last Appium version that did not exhibit the issue (if applicable): 1.4.16
* Desktop OS/version used to run Appium: Mac OS
* Node.js version (unless using Appium.app|exe): v5.7.0
* Mobile platform/version under test: iOS
* Real device or emulator/simulator: real 
* Appium CLI or Appium.app|exe: CLI

[Link to Appium logs](https://gist.github.com/HlebHalkouski/b745e1da71177a3f54b92e587ecd589a)

On javaclient@4.0.0 call method IOSDriver.swipe(200, 200, 0, 100, 3)

toejough:
Fixes #150 

swipe action itself taken from lib/devices/ios/uiauto/appium/app.js as removed by commit f4a1b6fcafc51ae830132b27ff0ed44b5e253865.

extra code to get the data for the swipe action taken from other sections of the existing gesture.js in appium-ios-driver.

This is running quite well in my app tests at the moment.

I've tried to match the code style with this PR, but I didn't see any official guidelines for this repo (though I'm happy to meet standards if they are pointed out).

It also doesn't have any added unit tests.  When I ran the existing tests locally, they passed, I think, but the E2E tests definitely didn't.  I'd be happy to add some more unit tests.

### DIFF
--- a/lib/commands/gesture.js
+++ b/lib/commands/gesture.js
@@ -114,6 +114,89 @@ helpers.handleTap = async function (gesture) {
   return await this.uiAutoClient.sendCommand(cmd);
 };
 
+helpers.isDrag = function (gestures) {
+  return (
+      gestures.length === 4 &&
+      gestures[0].action === 'press' &&
+      gestures[1].action === 'wait' &&
+      gestures[2].action === 'moveTo' &&
+      gestures[3].action === 'release'
+  );
+};
+
+helpers.getCoordinates = async function(gesture) {
+  let el = gesture.options.element;
+
+  // defaults
+  let coordinates = {x: 0, y: 0, areOffsets: false};
+
+  // figure out the element coordinates.
+  if (el) {
+    let command = `au.getElement('${el}').rect()`;
+    let rect = await this.uiAutoClient.sendCommand(command);
+    let pos = {x: rect.origin.x, y: rect.origin.y};
+    let size = {w: rect.size.width, h: rect.size.height};
+
+    // defaults
+    let offsetX = 0;
+    let offsetY = 0;
+
+    // get the real offsets
+    if (gesture.options.x || gesture.options.y) {
+      offsetX = (gesture.options.x || 0);
+      offsetY = (gesture.options.y || 0);
+    } else {
+      offsetX = (size.w / 2);
+      offsetY = (size.h / 2);
+    }
+
+    // apply the offsets
+    coordinates.x = pos.x + offsetX;
+    coordinates.y = pos.y + offsetY;
+  } else {
+    // moveTo coordinates are passed in as offsets
+    coordinates.areOffsets = (gesture.action === 'moveTo');
+    coordinates.x = (gesture.options.x || 0);
+    coordinates.y = (gesture.options.y || 0);
+  }
+  return coordinates;
+};
+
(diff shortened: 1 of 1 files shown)


imurchie:
Thanks for this!

You'll need to run `gulp eslint` and fix anything that comes up, or all the tests will fail.

toejough:
ah, will do.  sorry I missed that - it's been a while since I've written in js.

lib/commands/gesture.js
```diff
@@ -114,13 +114,108 @@ helpers.handleTap = async function (gesture) {
...
+    };
+  } else {
+    return secondCoordinates;
+  }
+};
+
+helpers.handleDrag = async function (gestures) {
+  // get gestures
+  let press = gestures[0];
+  this.validateGestureType(press, 'press');
```
imurchie: What is the point of these checks? This code is only ever called if the `isDrag` check has already passed, so these three checks are guaranteed to pass.

lib/commands/gesture.js
```diff
@@ -114,13 +114,108 @@ helpers.handleTap = async function (gesture) {
...
+  let wait = gestures[1];
+  this.validateGestureType(wait, 'wait');
+  let moveTo = gestures[2];
+  this.validateGestureType(moveTo, 'moveTo');
+
+  // get drag data
+  let pressCoordinates = await this.getCoordinates(press);
+  let duration = (parseInt(wait.options.ms) / 1000);
```
imurchie: `parseInt` needs a second parameter, `10`, otherwise the behavior is unpredictable.

```js
let duration = parseInt(wait.options.ms, 10) / 1000;
```

imurchie:
A couple of comments. But looks good overall! Thanks for taking this on!

lib/commands/gesture.js
toejough: they're just defensive - there's nothing stopping anyone from calling handleDrag with the wrong parameters later, and this makes sure that specifically, the gestures `handleDrag` actually uses are the correct gestures.

Could accomplish a similar thing with less code if `handleDrag` just called `isDrag(gestures)` at the top.  You'd get an error if there were greater/less than 4 gestures and the last one wasn't `release`.  I didn't do that because the function doesn't actually need those guarantees in order to generate the command it's supposed to generate.

toejough:
cool - I should be able to get the changes in today.  Do you want changes to the validation code in `handleDrag`?

imurchie:
I don't think it's necessary. Users don't call any of this code, so it is defending against other developers on the system.

imurchie:
This looks good! One last thing to do is to squash into a single commit. Then we can get to merging!

toejough:
squashed!  I just signed the contributor agreement from the main appium project, too.

I left the original commit message - let me know if it needs to be rewritten with any particular data/wording.

imurchie:
Thank you for doing this! It's great!

imurchie:
Published in `appium-ios-driver@1.12.11`.

toejough:
Whoohoo!  happy to have been able to contribute!

PR #164, opened 2016-07-29
Export commands

imurchie:
Allow other drivers to use the code for these commands. They are mixed into this driver, and exporting them allows them to be mixed into other drivers.

There may be some tweaks needed for particular commands, but this will begin the job.


jlipps:
👍

PR #166, opened 2016-08-08
Fix variable name for iosSdkVersion

imurchie:
Everywhere else (including where it is set) the variable is `this.iosSdkVersion` not `this.iOSSDKVersion`.


jlipps:
👍

PR #167, opened 2016-08-11
Include the context element when attempting to find a descendent via XPath

jdeff:
All requests to find elements from an element using the xpath strategy are currently failing. It turns out that we are never passing the context element to uiauto's "getElement(s)ByIndexPath" methods.

This PR includes both the fix for finding descendent elements using XPath and a corresponding e2e test.

### DIFF
--- a/lib/commands/find.js
+++ b/lib/commands/find.js
@@ -173,13 +173,23 @@ commands.findUIElementsByXpath = async function (selector, mult, context=null, c
   }
 
   // otherwise look up the actual element by its index path
-  let proxyCmd;
+  let methodName;
+  let methodArgs = [];
+
   if (!mult) {
-    proxyCmd = `au.getElementByIndexPath('${indexPaths[0]}')`;
+    methodName = "getElementByIndexPath";
+    methodArgs[0] = `'${indexPaths[0]}'`;
   } else {
-    let ipArrString = JSON.stringify(indexPaths);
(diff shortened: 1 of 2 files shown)


jlipps:
wow, this is a pretty big oversight. thank you. my only comment is that we prefer to use template literals instead of concatenation now; mind redoing in that style?

e.g.

```js
  methodArgs[1] = `au.getElement('${context}')`;
```

imurchie:
Nice catch! Thanks!

jdeff:
@jlipps, fixed.

imurchie:
Looks good to me! Thanks for catching this and fixing it!

imurchie:
Published in `appium-ios-driver@1.12.16`

jdeff:
So what's the preferred method of getting this fix? Is it just to wait until another version of Appium is release? I couldn't quite figure out how to get npm to give me what I wanted without either deleting the npm-shrinkwrap.json file and updating from within the appium module or running `npm install` with `--no-shrinkwrap`. I don't really want to either of those if there is an easy way to do what I want that I am just missing ...

jlipps:
Now what you'd need to do is, in your local checkout of Appium, `rm -rf node_modules && npm install`. That should bring in the new version of ios-driver.

Until we release a new version of appium it will not be included outside of running from source however.

PR #168, opened 2016-08-11
Send page load along to remote debugger

imurchie:
Small change to actually send along the page load timeout to where it matters. Tests are skipped because the page load timeout is not actually respected in `appium-remote-debugger`.


jlipps:
👍  however if we're going to commit a `skip` let's make sure to add a comment as to why

imurchie:
Haha! Yes. I added the comment to the version in `appium-xcuitest-driver` and not here.

PR #169, opened 2016-08-12
Fix page change logic

imurchie:
We never actually got into this page-load logic, since what we were comparing was wrong.


jlipps:
👍

PR #170, opened 2016-08-17
Extract settings and device functions so they can be used elsewhere

imurchie:
Pull settings code out into its own file and try to contain some of the functionality so it is not all methods on the IosDriver class.

This way it can be brought into other classes more easily.

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -5,7 +5,7 @@ import path from 'path';
 import _ from 'lodash';
 import B from 'bluebird';
 import { fs } from 'appium-support';
-import { getSimulator, getDeviceString } from 'appium-ios-simulator';
+import { getSimulator } from 'appium-ios-simulator';
 import { prepareBootstrap, UIAutoClient } from 'appium-uiauto';
 import { Instruments, utils as instrumentsUtils } from 'appium-instruments';
 import { retry, waitForCondition } from 'asyncbox';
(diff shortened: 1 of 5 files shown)


lib/driver.js
```diff
@@ -292,12 +293,10 @@ class IosDriver extends BaseDriver {
       }
     }
 
-    await this.checkPreferences();
-
     await this.runSimReset();
     await this.isolateSimDevice();
-    await this.setLocale();
-    await this.setPreferences();
+    await this.setLocale(this.sim);
```
jlipps: since we're pulling `this.sim` out, it makes me wonder if we can go whole hog and remove all instances of it; do we really use it anywhere other than in functions called by `startSimulator`?

lib/settings.js
```diff
@@ -0,0 +1,156 @@
...
+    if (_.has(opts, setting)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// pass in the simulator so that other systems that use the function can supply
+// whatever they have
+async function setLocale (sim) {
```
jlipps: maybe we do or don't want to do this now, but we could refactor this so that it took parameters like opts and whether something is a real device, so that this becomes a self-contained function which doesn't need to be added to a prototype in order to be useful.

jlipps:
made a couple comments. if we don't want to address them now, we could add a `TODO` to point out where we can make our code less coupled.

lib/settings.js
imurchie: I would love for most of these functions (indeed, most of the functions on the driver). I'm working on the reset logic, and will try to see if I can make them as functional as possible.

PR #174, opened 2016-08-24
export default server capabilities

jlipps:
so we can use them in xcuitest too.

should assist with https://github.com/appium/appium/issues/6771

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -29,7 +29,7 @@ let iDevice = function (...args) {
   return promisified;
 };
 
-let serverCapabilities = {
+const defaultServerCaps = {
   webStorageEnabled: false,
   locationContextEnabled: false,
   browserName: '',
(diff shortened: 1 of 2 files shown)


lib/driver.js
```diff
@@ -43,7 +43,6 @@ let serverCapabilities = {
 class IosDriver extends BaseDriver {
   resetIos () {
     this.appExt = ".app";
-    _.defaults(this.opts, serverCapabilities);
```
jlipps: i couldn't think of any reason we actually wanted these on `this.opts` other than to return them, which we do by extending down below. so removing for sanity. but it's possible these are needed somehow. cc @imurchie

imurchie: Yeah, I don't think we make use of these.

imurchie:
Looks good to me! It seems to have gotten through the build once, during a window when Travis worked.

I'm going to merge so we can get the `appium-xcuitest-driver` fix through.

imurchie:
Published in `appium-ios-driver@1.12.20`

PR #177, opened 2016-08-30
Fix the this.opts.curOrientation 

bigfei:
Since uiAutoClient.sendCommand is promisified, there is no need to check res.status and use the result directly.

### DIFF
--- a/lib/commands/orientation.js
+++ b/lib/commands/orientation.js
@@ -10,7 +10,7 @@ commands.getOrientation = async function () {
     logger.debug(`Setting internal orientation to '${res}'`);
     this.opts.curOrientation = res;
     return res;
-  } catch (err){
+  } catch (err) {
     logger.error(`Device orientation is not supported.`);
     throw err;
   }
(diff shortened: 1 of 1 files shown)


lib/commands/orientation.js
```diff
@@ -4,24 +4,29 @@ import logger from '../logger';
...
+  try {
+    let res = await this.uiAutoClient.sendCommand('au.getScreenOrientation()');
     // keep track of orientation for our own purposes
-    let orientation = res.value.toUpperCase();
-    logger.debug(`Setting internal orientation to '${orientation}'`);
-    this.opts.curOrientation = orientation;
+    logger.debug(`Setting internal orientation to '${res}'`);
+    this.opts.curOrientation = res;
+    return res;
+  } catch (err){
```
jlipps: can you add a space here so that it is:

```js
} catch (err) {
```
?

jlipps:
serious oversight! thank you for the fix. i asked for one style change then this should be good to merge.

jlipps:
published as 1.12.23

PR #184, opened 2016-09-15
Stop uiauto and instruments after Safari Browser launching

nghiadhd:
We fixed https://github.com/appium/appium-instruments/issues/79

The root cause is after the instruments start SafariLauncher to launch the Safari Browser, it still exist and waiting for next command with timeout (3600s)
https://github.com/appium/appium-uiauto/blob/master/uiauto/lib/commands.js#L79

So if we do long test script (more than 1 hour) the timeout was reached and it throws "Abnormal Instruments termination!" and then Appium ends the session
We can reproduce (100%) with this test https://gist.github.com/nghiadhd/ae108085df5c16781271f34af0038589#file-long-run-js

We expect the instrument should be shutdown after Safari Browser launching

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -149,23 +149,30 @@ class IosDriver extends BaseDriver {
     return [sessionId, this.caps];
   }
 
-  async stop () {
-    this.ready = false;
-
+  async stopUIAutoClient () {
     if (this.uiAutoClient) {
       await this.uiAutoClient.shutdown();
+      this.uiAutoClient = null;
     }
+  }
 
+  async stopInstruments () {
     if (this.instruments) {
       try {
         await this.instruments.shutdown();
       } catch (err) {
         logger.error(`Instruments didn't shut down. ${err}`);
       }
+      this.instruments = null;
     }
+  }
 
-    this.uiAutoClient = null;
-    this.instruments = null;
+  async stop () {
+    this.ready = false;
+
+    await this.stopUIAutoClient();
+    await this.stopInstruments();
+
     this.realDevice = null;
 
     // postcleanup
@@ -497,6 +504,12 @@ class IosDriver extends BaseDriver {
       }
       logger.debug('Waiting for initial webview');
       await this.navToInitialWebview();
+      if (this.isRealDevice()) {
+        // we don't need UIAutoClient and Instruments any more (on real device),
+        // should stop them to avoid 'Abnormal Instruments error' due to the command timeout.
+        await this.stopUIAutoClient();
+        await this.stopInstruments();
+      }
       condFn = async () => {
         return true;
       };


lib/driver.js
```diff
@@ -497,6 +508,10 @@ class IosDriver extends BaseDriver {
       }
       logger.debug('Waiting for initial webview');
       await this.navToInitialWebview();
+      // we don't need UIAutoClient and Instruments any more,
+      // should stop them to avoid 'Abnormal Instruments error' due to the command timeout.
+      await this.stopUIAutoClient();
```
jlipps: i don't think we want to do this except in the case of a real device. With a sim, we still have a good instruments session!

nghiadhd: Hi @jlipps 

I didn't test on Simulator because I only support real device. But if we still use the instrument to launch Safari Browser may the bug still affect too. Because in long time (more than 1hour) the instrument doesn't get any command from uiAutoClient and may it got timeout.

jlipps: You may be right about that. However, the cost is too high. With automating safari on a simulator, we have access to the full UI of Safari via UIAutomation. If we stop the uiauto client we will no longer have that access. It's very important that we keep it. So I would rather risk a bug that only pops up after an hour of running an app continuously than remove functionality which many people currently rely on.

nghiadhd: Got it @jlipps 
I'll do the change request and update soon

nghiadhd: Hi @jlipps 

I see we got failed when running with CI.
But "dbug Instruments [INST] 2016-09-15 15:18:19 +0000 Debug: Unlock failed. Retrying up to 1 more time(s)." seem doesn't relate with my code. So now I would like to run it in my local. How can I do!

nghiadhd:
@jlipps 

Could you help me the guide to run e2e for appium-ios-driver? I run the commands in README but it doesn't launch the simulator. My environment is iOS 9.3, Xcode 7.3.1
I would like to run e2d in my local because I see the failed test in log file seem doesn't relate with my code. If don't have this patch, I cannot pass the test in my side

imurchie:
You should be able to run `_FORCE_LOGS=1 gulp e2e-test`to run all the tests.

nghiadhd:
@jlipps 

I can run e2e in my environment and see that the failed cause by this code line
let allowance = (env.IOS7 || env.IOS8 || env.IOS9) ? 10 : 2;

I see allowance always is 2, 
Deeply debug and I see that env.IOS7 or env.IOS8 or env.IOS9 is null while environment is setting to use iOS 9.3 & Xcode 7.3.1

```
info iOS IosDriver server listening on http://localhost:4723
dbug iOS caps --> { app: '/Volumes/Data/Projects/appium-ios-driver/node_modules/ios-uicatalog/build/Release-iphonesimulator/UICatalog-iphonesimulator.app',
dbug iOS   noReset: true,
dbug iOS   browserName: '',
dbug iOS   deviceName: 'iPhone 6',
dbug iOS   launchTimeout: 60000,
dbug iOS   platformName: 'iOS' }
dbug iOS opts --> { port: 4723,
dbug iOS   callbackAddress: 'localhost',
dbug iOS   tmpDir: '/var/folders/my/chbbx01x2x7d9h1nq2qxbqy40000gn/T' }
dbug iOS remainingAttempts --> 3
info BaseDriver Session created with session id: 4c028a54-8802-426c-b352-5b79c800c737
dbug iOS Xcode version set to 7.3.1
dbug iOS Not auto-detecting udid.
info BaseDriver Using local app '/Volumes/Data/Projects/appium-ios-driver/node_modules/ios-uicatalog/build/Release-iphonesimulator/UICatalog-iphonesimulator.app'
dbug iOS Removing any remaining instruments sockets
dbug iOS Cleaned up instruments socket /var/folders/my/chbbx01x2x7d9h1nq2qxbqy40000gn/T/instruments_sock
dbug iOS Setting iOS SDK Version
dbug iOS iOS SDK Version set to 9.3
```

I tried to set allowance is 10 and now I can pass the "lock device" test.
So could you double check in your side?

jlipps:
@imurchie maybe the env var isnt getting set correctly for ios9?

imurchie:
Yes. The command I gave didn't have the DEVICE set, so nothing will be there. 

It needs to be have `DEVICE=ios93` in it.

But that does not answer why the job is failing in Travis.

nghiadhd:
@imurchie 

I've just run with (I re-config to run lock-device-specs.js)
`_FORCE_LOGS=1 DEVICE=ios93 gulp e2e-test`

2 passing (44s)

Log: https://gist.github.com/nghiadhd/fe10fe766334895fc747dffc1f447047#file-device_lock-log

nghiadhd:
@imurchie 
Please review this PR, Now All checks have passed

nghiadhd:
@imurchie and @jlipps 

Does this PR has any issue? if not, please help me merge it

tin-tran-logigear:
Thank @nghiadhd ! It's helpful. But this issue hasn't been merged to master.

jlipps:
Hi @nghiadhd sorry I didn't see this before, but the code looks good to me. @imurchie can we merge?

imurchie:
@jlipps I thought we were blocked on your review/reservations (https://github.com/appium/appium-ios-driver/pull/184#discussion_r78977685).

tin-tran-logigear:
Hi @jlipps Can you check this again?

jlipps:
My reservations have been addressed with the changes. @nghiadhd can you rebase to deal with conflicts? Then we can merge.

nghiadhd:
@jlipps 

After do some more test, I see my fix has issue
If you do on all actions in browser, it should be fine but when you do any action relate to system such as change orientation, now it's problem. The instrument has stopped and appium-ios-driver throws the "undefined.*" exception and the test has stopped

May we need to investigate how to use ping action instead set timeout is 3600s

jlipps:
@nghiadhd good points. I'll close this until we can come up with a better idea.

PR #188, opened 2016-10-12
Add clearSystemFiles cap to delete cache and logs at the end of the session

imurchie:
Add a capability, `clearLogs`, to indicate that the cache and log files should be deleted at the end of the session.

Address https://github.com/appium/appium/issues/6952

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -233,7 +233,29 @@ function unwrapEl (el) {
   return el;
 }
 
+async function clearLogs (locations) {
+  logger.debug('Clearing log files');
+  for (let location of locations) {
+    if (await fs.exists(location)) {
+      let size;
+      try {
+        let {stdout} = await exec('du', ['-sh', location]);
+        size = stdout.trim().split(/\s+/)[0];
+      } catch (ign) {}
+      try {
+        /* jshint ignore:start */
+        logger.debug(`Deleting '${location}'. ${size ? `Freeing ${size}.` : ''}`);
+        /* jshint ignore:end */
+        await fs.rimraf(location);
+      } catch (err) {
+        logger.warn(`Unable to delete '${location}': ${err.message}`);
+      }
+    }
+  }
+  logger.debug('Finished clearing log files');
+}
+
 export default { rootDir, removeInstrumentsSocket, getAndCheckXcodeVersion, prepareIosOpts,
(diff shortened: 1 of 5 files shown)


lib/utils.js
```diff
@@ -233,7 +233,34 @@ function unwrapEl (el) {
   return el;
 }
 
+async function clearLogs (locations, clear = true) {
```
jlipps: seems a little odd to me to have a function which takes as a boolean parameter whether or not that function is going to do anything. what about instead, up where we call this, doing:

```js
if (!!this.opts.clearLog) {
  await utils.clearLogs(LOG_LOCATIONS);
} else {
  log.info("Not clearing logfiles");
}
```

imurchie: Fair enough!

imurchie:
Changed capability to `clearSystemFiles` as in Android it can be implemented to delete temporary copies of the apk, etc (see, for instance, https://github.com/appium/appium/issues/5156).

vishwathma:
This is the sample of the hard disk space consumption after running few number of appium tests:
Before running any test case, this was the memory:                                                                            
Node 1 - 19.91 GB   Node 2 - 9.78 GB    Node 3 - 10.23 GB

After running around 40 test cases without closing the appium session:
 Node 1 - 12.93  GB   Node 2 - 4.22 GB    Node 3 - 6.13 GB
                   
After stopping and starting the appium server in each of the nodes:
Node 1 - 21.75 GB   Node 2 -14.52 GB    Node 3 - 12.3 GB

@imurchie, I had a question for you, if we run the test cases in the same session(without ending the session), the hard disk memory gets consumed rapidly after the execution of each test case. What are your thoughts on how to overcome this problem?

imurchie:
The files taking up space are used by Apple's automation system. I don't think there is a reliable way to remove any without destroying a test run. So mid-session cleanup is probably not going to work. It's also not clear when Appium would do that, given that it doesn't know anything about the tests being run.

vishwathma:
@imurchie , this seems to be the same issue as https://github.com/calabash/run_loop/pull/306 and it is specified this has been fixed. Isn't it?

imurchie:
It looks to be the same thing as a little part of that PR, yes. I'm not sure what you mean by "it is specified this has been fixed".

PR #190, opened 2016-11-01
Fix reset on real device

imurchie:
When these methods were pulled off the object, this `this` slipped through.

Fixes https://github.com/appium/appium/issues/7092


imurchie:
Published in `appium-ios-driver@1.12.31`.

PR #196, opened 2016-11-23
Accepts customSSLCert as a capability

dpgraham:
-Updated driver.js to installSSLCert when it's using an IOS Simulator and customSSLCert was provided.
-Updated appium-ios-simulator to 1.9.0 which has installSSLCert and uninstallSSLCert methods
-Wrote a spec that sets up a temporary https site with a random PEM certificate to test that it can open said site
-Make customSSLCert part of desired caps

### DIFF
--- a/lib/driver.js
+++ b/lib/driver.js
@@ -5,7 +5,7 @@ import path from 'path';
 import _ from 'lodash';
 import B from 'bluebird';
 import { fs } from 'appium-support';
-import { getSimulator } from 'appium-ios-simulator';
+import { getSimulator, installSSLCert, uninstallSSLCert } from 'appium-ios-simulator';
 import { prepareBootstrap, UIAutoClient } from 'appium-uiauto';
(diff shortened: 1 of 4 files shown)


lib/driver.js
```diff
@@ -174,6 +174,10 @@ class IosDriver extends BaseDriver {
       }
     }
 
+    if (this.caps.customSSLCert && !this.isRealDevice()) {
+      uninstallSSLCert(this.caps.customSSLCert, this.iosSimUdid);
```
jlipps: doesn't this need to be `await`ed?

lib/driver.js
```diff
@@ -314,6 +318,10 @@ class IosDriver extends BaseDriver {
       }
     }
 
+    if (this.caps.customSSLCert && !this.isRealDevice()) {
+      installSSLCert(this.caps.customSSLCert, iosSimUdid);
```
jlipps: same here

PR #197, opened 2016-11-25
Only call this.remote.pageLoad() when necessary in the appium-ios-driver

cschabetsberger-micro-focus:
Fixed the check of needsPageLoad: this.contexts contains elements of the form &lt;appId&gt;.&lt;pageId&gt; and therefore we need to make sure that contextArray contains elements of the same structure

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -299,13 +299,13 @@ extensions.onPageChange = async function (pageChangeNotification) {
       let needsPageLoad = (() => {
         let item = (arr) => {
           return _.filter(arr, (obj) => {
-            return obj.id === this.curContext;
+            return obj === this.curContext;
           })[0];
         };
 
         // need to map the page ids to context ids
         let contextArray = _.map(pageArray, (arr) => {
-          return {id: `${appIdKey}.${arr.id}`};
+          return `${appIdKey}.${arr.id}`;
         });
         return !_.isEqual(item(this.contexts), item(contextArray));
       })();


lib/commands/context.js
```diff
@@ -299,13 +299,13 @@ extensions.onPageChange = async function (pageChangeNotification) {
       let needsPageLoad = (() => {
         let item = (arr) => {
           return _.filter(arr, (obj) => {
-            return obj.id === this.curContext;
```
jlipps: @cschabetsberger-micro-focus I think this change makes sense. However I'm a bit stuck trying to understand the existing code. With your changes, how I read this section:

```js
        let item = (arr) => {
          return _.filter(arr, (obj) => {
            return obj === this.curContext;
          })[0];
        };
```

Is that it will either result in `this.curContext` or `undefined`, based on whether the array passed in contains something equivalent to `this.curContext`.

So, calling `item(this.contexts)` below seems like a roundabout way to just reference `this.curContext`, no? Surely `this.contexts` would always contain `this.curContext`?

In any case, if I'm reading this right, we're basically just checking whether the new pageArray contains `this.curContext`, and returning `false` if it does, to say we don't need a page change. If that's the case, sounds like we can get rid of that block I mentioned in favor of something more transparent?

@imurchie should probably weigh in on this.

cschabetsberger-micro-focus: I think it can theoretically happen that this.curContext has a certain value and then this.contexts changes during runtime (for example, because getContexts() is called and the context from this.curContext is not available anymore). If this happens, item(this.contexts) would return undefined and a pageLoad would only be triggered if contextArray contains this.curContext. However, I don't know if this is necessary as I don't know the code well enough.

jlipps:
@imurchie can you comment on this simplification?

NiNi1995:
C:\Users\Administrator>appium-doctor
module.js:471
    throw err;
    ^

Error: Cannot find module '../lib/doctor/ios.js'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (C:\Program Files (x86)\Appium\node_modules\appium\bin\appium-doctor.js:4:49)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)

imurchie:
Closing as implemented in https://github.com/appium/appium-ios-driver/pull/250 and available in `appium-ios-driver@1.27.4`

PR #198, opened 2016-12-02
Uninstall ssl fix 

dpgraham:
Fixed problem in Travis https://travis-ci.org/appium/appium-ios-driver/jobs/178882412

I couldn't reproduce the issue locally but I think what's happening is that it was testing on real device, which wasn't installing the SSL cert and then was attempting to uninstall it and throwing an undefined exception. This fix does a check to make sure this.caps is there. Also added the 'await' for installSSLCert and uninstallSSLCert.

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -81,21 +81,23 @@ commands.lock = async function (secs) {
 };
 
 commands.closeApp = async function () {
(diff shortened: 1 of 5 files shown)


jlipps:
awesome, can you squash?

jlipps:
publishing in 1.13.1

PR #201, opened 2016-12-12
Fix the build: use correct udid for uninstall ssl cert

imurchie:
Fix the `udid` being sent into the uninstall ssl command.

Also quarantine the ssl test in Travis, since it is broken there.


dpgraham:
Thanks for fixing this @imurchie

PR #206, opened 2016-12-16
Move appium-ios-log into driver

imurchie:
First part of https://github.com/appium/appium/issues/7467


imurchie:
Published in `appium-ios-driver@1.15.0`

PR #213, opened 2017-02-23
changing 127.0.0.1 to appium's listening address

alizelzele:
While testing safari on ios, we found that when initializing driver appium tries to open url: `http://127.0.0.1:4723/welcome`. It is ok if you are listening on default address (0.0.0.0), but if you listen on another ip by using `-a` parameter, appium fails to open this url.

Changes in this PR solve this issue.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -396,7 +396,8 @@ async function openNewPage () {
 }
 
 extensions.typeAndNavToUrl = async function () {
-  this.setCurrentUrl(this.caps.safariInitialUrl || `http://127.0.0.1:${this.opts.port}/welcome`);
+  let address = this.opts.address ? this.opts.address : '127.0.0.1';
+  this.setCurrentUrl(this.caps.safariInitialUrl || `http://${address}:${this.opts.port}/welcome`);
 
   let tries = 0;
   const MAX_TRIES = 2;


lib/commands/context.js
```diff
@@ -396,7 +396,8 @@ async function openNewPage () {
 }
 
 extensions.typeAndNavToUrl = async function () {
-  this.setCurrentUrl(this.caps.safariInitialUrl || `http://127.0.0.1:${this.opts.port}/welcome`);
+  let address = (this.opts.address === '0.0.0.0') ? '127.0.0.1' : this.opts.address;
```
imurchie: Does this not work just keeping `0.0.0.0` for the base case? It seems to on my machine.

jlipps: i ran into an issue with trying to actually access `0.0.0.0` on windows, so i think it fails there. but this is for ios and i think it is successful. probably works if we just do `let address = this.opts.address`?

alizelzele: Silly me. `0.0.0.0` works on unix, but failed on windows. I actually add this line to support windows! (facepalm) :rofl:

Will update the PR asap.

imurchie:
The address is getting to be "undefined" when the driver is run alone.

```
dbug RemoteDebugger Received app, but expected url ('http://undefined:4723/welcome') was not found. Trying again.
```

The line will have to include checking that there is indeed a value in `this.opts.address`.

alizelzele:
@imurchie just saw your comment. on my system it was always returning 0.0.0.0. I will check for it not being `undefinied`, ASAP :)

imurchie:
Thanks for this!

PR #215, opened 2017-02-28
Internally control ios_webkit_debug_proxy

dpgraham:
* Step one for implementing https://github.com/appium/appium/issues/4454
* Uses teen_process (promise implementation of 'spawn') to open the ios_webkit_debug_proxy from within ios-driver
* Restarts it if it stops abruptly (limit 10 retries)
* Added dcap called 'startIWDP' which tells the driver to use ios_webkit_debug_proxy internally
* Next step is to update xcuitest-driver to use this capability and to run functional tests on a real device to verify that it's doing it's job


lib/driver.js
```diff
@@ -468,6 +470,12 @@ class IosDriver extends BaseDriver {
     }
   }
 
+  async startIWDP () {
+    if (this.opts.runIWDP) {
```
jlipps: you called the cap `startIWDP` but here it is `runIWDP`?

lib/driver.js
```diff
@@ -468,6 +470,12 @@ class IosDriver extends BaseDriver {
...
+      IWDP.start(this.opts.udid);
```
jlipps: this should be `await`ed

lib/iwdp.js
```diff
@@ -0,0 +1,78 @@
+import { SubProcess } from 'teen_process';
+import EventEmitter from 'events';
+import { execSync } from 'child_process';
+import Promise from 'bluebird';
+import logger from './logger';
+
+const request = Promise.promisify(require('request'));
```
jlipps: typically we use the `request-promise` package instead of promisifying request ourselves

lib/iwdp.js
```diff
@@ -0,0 +1,78 @@
...
+
+  createIWDPProcess () {
+    let process = new SubProcess(IWDP_CMD, ['--config', `${this.udid}:${this.port},:${this.port + 1}-${this.port + 101}`]); // (see https://github.com/google/ios-webkit-debug-proxy for reference)// If the process exits and the exit wasn't requested by the API, restart it
+    process.on('exit', () => this.onExit());
+    return process;
+  }
+
+  async onExit () {
+    if (!this.exitRequested) {
+      delete this.process;
```
jlipps: not sure what `delete`ing does that simply reassigning doesn't; pretty sure JS gc will handle it the same? but maybe i'm wrong

lib/iwdp.js
```diff
@@ -0,0 +1,78 @@
...
+      return this.emit('error', new Error(`Failed to start IWDP server. Max retry attempts ${MAX_RETRIES} reached`));
+    }
+    this.exitRequested = false;
+    await this.process.start(0);
+
+    // Don't return until we confirm that the endpoint is accessible
+    const TIMEOUT = 10000;
+    const ENDPOINT_CHECK_DELAY = 300;
+    let totalDelay = 0;
+    while (totalDelay < TIMEOUT) {
```
jlipps: rather than constructing retry logic ourselves, what about asyncbox's `retryInterval`?

jlipps: with that we don't need to keep track of number of attempts, timeout, delay, etc...

lib/iwdp.js
```diff
@@ -0,0 +1,78 @@
...
+  async stop () {
+    this.attempts = 0;
+    this.exitRequested = true;
+    return await this.process.stop();
+  }
+
+	/**
+	 * Is 'ios_webkit_debug_proxy' available?
+	 */
+  static isSupported () {
```
jlipps: let's not do it with `execSync`; we have an appium-support extension to `fs` we should use, and this method should be async.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,76 @@
...
+
+  it('should not keep running after stop is called', async function () {
+    await iwdpInstance.start();
+    await iwdpInstance.stop();
+    request(iwdpInstance.endpoint).should.be.rejected;
+  });
+
+  it('should still start IWDP server if one is started on a different port', async function() {
+    let process = new SubProcess('ios_webkit_debug_proxy', ['c', 'null:12345']);
+    await process.start();
```
jlipps: probably want to clean up `process` here after we use it

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,76 @@
...
+      await iwdpInstance.stop();
+    } catch (ign) { }
+  });
+
+  it('should detect that IWDP is supported on this machine', function () {
+    expect(IWDP.isSupported());
+  });
+
+  it('should start IWDP and be able to access the main page', async function () {
+    await iwdpInstance.start();
```
imurchie: For this to work you'll need to update `.travis.yml` to install the webkit proxy. The Travis VM has homebrew on it, so it should be pretty easy.

lib/iwdp.js
```diff
@@ -0,0 +1,78 @@
...
+import logger from './logger';
+
+const request = Promise.promisify(require('request'));
+
+const IWDP_CMD = 'ios_webkit_debug_proxy';
+const MAX_RETRIES = 10;
+
+class IWDP extends EventEmitter {
+
+  constructor (udid = null) {
```
imurchie: We need to be able to support arbitrary ports for this, passed to the ios driver through `opts. webkitDebugProxyPort` (see https://github.com/appium/appium/blob/master/lib/parser.js#L345).

lib/iwdp.js
imurchie: 💯

lib/iwdp.js
dpgraham: Isn't the 'webkitDebugProxyPort' configuration instructing Appium which port `ios_webkit_debug_proxy` is opened at?

I think in this case, because the driver is opening ios_webkit_debug_proxy we want to set it to a fixed port.

imurchie: I still think we ought to have it configurable. I'm sure there is a valid use case for wanting things to run on a specific port.

lib/driver.js
```diff
@@ -468,6 +470,12 @@ class IosDriver extends BaseDriver {
     }
   }
 
+  async startIWDP () {
+    if (this.opts.startIWDP) {
+      await IWDP.start(this.opts.udid);
```
imurchie: Does this work? Treating this as a static method? We don't want to be able to keep track of the object? Close iwdp when done?

dpgraham: Yeah that's a mistake. I think that method used to be static and now it's a class. Good catch.

lib/iwdp.js
dpgraham: Should we use the same flag or should I add a new flag on top of the current one?

jlipps: pretty sure we have a cap for iwdPort already, and we should just use that

lib/iwdp.js
```diff
@@ -0,0 +1,81 @@
...
+    super();
+    this.udid = udid;
+    this.attempts = 0;
+    this.port = 27753;
```
jlipps: let's pull out default constants like this to constant vars at the top of the file (if we still want this defined in this file at all to begin with)

lib/iwdp.js
```diff
@@ -0,0 +1,81 @@
...
+	/**
+	 * Is 'ios_webkit_debug_proxy' available?
+	 */
+  async isSupported () {
+    if (typeof(this.supported) !== 'undefined') {
+      return this.supported;
+    }
+
+    try {
+      await exec('which', [IWDP_CMD]);
```
jlipps: i still think we shouldn't run the system `which` but rather use the library function available in `appium-support`

lib/iwdp.js
imurchie: Same one ought to do.

lib/iwdp.js
```diff
@@ -0,0 +1,81 @@
...
+
+  async onExit () {
+    if (!this.exitRequested) {
+      this.process = this.createIWDPProcess();
+      await this.start(); 
+    }
+  }
+
+  async start () {
+    if (++this.attempts > MAX_RETRIES) {
```
jlipps: i don't see us attempting any retries at all, so why still have this block?

lib/iwdp.js
dpgraham: I found the method you're talking about. It wasn't in the appium-support readme.md so I made a PR: https://github.com/appium/appium-support/pull/46

lib/iwdp.js
dpgraham: When the process exits, `onExit` is called. If the 'exit' is unexpected (ie: it wasn't called in iwdp.stop method) then we call start again (`await this.start()`). If it does this 10 times then it emits an 'error' event and stops retrying.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,82 @@
+// transpile:mocha
+import chai from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+import IWDP from '../../lib/iwdp';
+import { SubProcess } from 'teen_process';
+import request from 'request-promise';
+import Promise from 'bluebird';
```
imurchie: We have generally been using `B` as the import for `bluebird`, mostly to distinguish it from the vanilla `Promise` implementation, which has different capabilities.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  it('should fail after reaching max retries', async (done) => {
+    await iwdpInstance.start();
+    let retries = 0;
+
+    // It should give up restarting after 10 failed attempts
+    iwdpInstance.on('error', () => {
+      expect(retries).to.equal(10);
+      done();
```
imurchie: For some reason this offends mocha on Travis (`Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.`). This is because of the `chai.use(chaiAsPromised)` directive, and then the `done` callback.

Something like this is needed:
```js
  it.only('should fail after reaching max retries', async () => {
    await iwdpInstance.start();
    let retries = 0;

    // It should give up restarting after 10 failed attempts
    let errorPromise = new Promise((resolve) => {
      iwdpInstance.on('error', () => {
        expect(retries).to.equal(10);
        resolve();
      });
    });

    // Keep stopping the process after it has been started
    while (++retries <= 10) {
      let promise = new Promise((resolve) => {
        iwdpInstance.once('start', () => {
          resolve();
        });
      });
      await iwdpInstance.process.stop();
      await Promise.any([promise, errorPromise]);
    }

    await errorPromise;
  });
```

dpgraham: Yeah I'm going to do it this way.

imurchie: 👍

lib/iwdp.js
jlipps: Still a little confused here. It seems that we could encounter an exit event in 3 situations:

1. While we're waiting for IWDP to start initially
2. When we're intentionally stopping IWDP
3. When IWDP exits at some other point in the middle of use.

We're already handling case 1 in the retryInterval (if IWDP doesn't start, we won't be able to request to it)

In case 2, we don't care about hearing when it exits.

In case 3, why would we want to limit the number of times we restart IWDP in the middle of a test?

test/unit/iwdp-specs.js
jlipps: Yeah, since we're using async functions, we never have reason to use `done()` in mocha tests.

dpgraham: I tried Isaac's way and for some inexplicable reason it wasn't resolving the promise so I did my own hack fix of making the function synchronous and then calling an async function from within it.

lib/iwdp.js
dpgraham: Case 2: When we intentionally stop IWDP a flag called `exitRequested` is set to `true` and then `process.stop()` is called which triggers `exit`. The callback for `exit` checks if `exitRequested === true` and if it does, it won't try restarting it.

Case 3: If the process is broken somehow and it it keeps exiting and we keep attempting to restart it, the tests will keep failing and the user won't know why. If we shut it down after 10 retries (10 was chosen arbitrarily, I could change that) then an error will be thrown and the user will be given logging information that explains the problem.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,86 @@
...
+
+      // Keep stopping the process after it has been started
+      while (++retries <= 10) {
+        let promise = new Promise((resolve) => {
+          iwdpInstance.once('start', () => {
+            resolve();
+          });
+        });
+        await iwdpInstance.process.stop();
+        await promise;
```
imurchie: This promise will hang forever, since it is never resolved on the 10th iteration.

dpgraham: Yep. That was the problem. Good catch.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,86 @@
...
+  it('should restart after the process is stopped abruptly', async function () {
+    await iwdpInstance.start();
+    await iwdpInstance.process.stop();
+    await new Promise((resolve) => {
+      iwdpInstance.once('start', resolve);
+    });
+    await request(iwdpInstance.endpoint).should.eventually.have.string('<html');
+  });
+
+  it('should fail after reaching max retries', (done) => {
```
dpgraham: I changed it over to not use `done` anymore.

test/unit/iwdp-specs.js
dpgraham: Done.

test/unit/iwdp-specs.js
```diff
@@ -0,0 +1,88 @@
...
+
+    // Keep stopping the process after it has been started
+    while (++retries <= 10) {
+      let promise = new B((resolve) => {
+        iwdpInstance.once('start', () => {
+          resolve();
+        });
+      });
+      await iwdpInstance.process.stop();
+      if (retries < 10) {
```
imurchie: This was why I was suggesting using `B.any([promise, errorPromise])`, so that either of them could be resolved.

dpgraham: Yeah I missed that part.

lib/iwdp.js
jlipps: ok, cool. i get it now.

lib/driver.js
```diff
@@ -18,6 +18,7 @@ import { setLocale, setPreferences } from './settings';
 import { runSimulatorReset, isolateSimulatorDevice, checkSimulatorAvailable,
          moveBuiltInApp, getAdjustedDeviceName, endSimulator, runRealDeviceReset } from './device';
 import { version } from '../../package.json'; // eslint-disable-line import/no-unresolved
+import { IWDP } from './iwdp';
```
jlipps: actually, we should probably have ios-driver export IWDP so that it can be used by xcuitest-driver, unless it simply calls ios-driver's `startRealDevice` method and `stop` method without modifications (which i don't think is the case)

imurchie: Yes. That is not the case. The class ought to be exported.

dpgraham: I just pushed a commit that exports the IWDP.

The next step, once this PR is closed, will be testing it in xcuitest-driver

dpgraham:
Is there anything left to remediate on this?

imurchie:
Nope I think we're good.

imurchie:
Published in `appium-ios-driver@1.17.0`.

dpgraham:
Thanks Isaac

PR #221, opened 2017-03-09
Fixed defect in retryInterval, improved logging

dpgraham:
* Retrying endpoints works when hitting localhost:27754 which is the endpoint for device. localhost:27753 is flakey
* Added logging for IWDP and log that says when IWDP starts
* Calls process.stop() on IWDP if it fails, just in case the process actually started and we don't want to leave it running
* Removed startIWDP capability (that goes in XCUITestDriver)

(please reveiw @imurchie @jlipps)

### DIFF
--- a/lib/desired-caps.js
+++ b/lib/desired-caps.js
@@ -96,9 +96,6 @@ const desiredCapConstraints = {
   customSSLCert: {
     isString: true
   },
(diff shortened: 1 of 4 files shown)


dpgraham:
Published to 1.17.1

PR #222, opened 2017-03-22
Remove date check for log output on real device

imurchie:
For real devices we check each log line to see if it is after the current server time, in order to make sure we aren't displaying logs before the current run. This means that if the device time happens to be earlier than the server time, no logs will _ever_ be displayed.

A bit of experimentation showed that `idevicesyslog` gives at most 50 (and normally closer to 10-20) lines before we attach, so the output is negligible (especially given the volume of logs that are generated by the device).

Related to https://github.com/appium/appium/issues/8068


test/e2e/safari/webview/special-caps-specs.js
```diff
@@ -1,50 +0,0 @@
-import desired from './desired';
```
jlipps: what's the rationale for removing these tests?

imurchie: I just moved them. The Safari tests have been epically flakey in this driver, and I can't figure out why. But since the tests themselves are weird, I though it might be the order.

PR #224, opened 2017-04-07
Handle native context in setUrl

sa-spag:
In order to be able to open deeplinks on iOS, I implemented a solution using `xcrun simctl openurl` command.


lib/commands/general.js
```diff
@@ -188,6 +188,19 @@ commands.getStrings = async function (language, stringFile = null) {
   return strings;
 };
 
+commands.setUrl = async function (url) {
+  log.debug(`Attempting to set url '${url}'`);
+  if (!this.isWebContext()) {
+    // use xcrun to open deeplink
+    await exec('xcrun', ['simctl', 'openurl', this.opts.udid || this.sim.udid, url]);
```
imurchie: There is a method on `node-simctl` for this command: https://github.com/appium/node-simctl/blob/master/lib/simctl.js#L67

sa-spag:
I guess we need to create a deeplink on https://github.com/appium/ios-test-app (something like `testapp://sum?a=(integer)&b=(integer)`) to test that.

jlipps:
@sa-spag yeah this is great. do you know how to create a deeplink scheme? would you mind doing it for TestApp?

sa-spag:
@jlipps Yep, that's why I updated it using a build created from https://github.com/appium/ios-test-app/pull/16. Perhaps we could add https://github.com/appium/ios-test-app as a dependency and change the path [here](https://github.com/appium/appium-ios-driver/blob/master/test/e2e/testapp/desired.js#L6)?

jlipps:
@sa-spag I think @imurchie made it this way because we were having trouble building the test app on CI, and so including it pre-built was easier. anyway, for now you've got the change here as well which is what's important. we can figure out what else to do later.

imurchie:
We can't just have the test app as a dev dependency because we need to run this repo in Travis using Xcope 7.3 and a recent change to the test app makes it impossible to build below 8.

imurchie:
Published in `appium-ios-driver@1.19.0`.

PR #227, opened 2017-04-17
Move IWDP tests from unit to e2e

dpgraham:
* Also added test that checks that calls to IWDP are rejected

(related to https://github.com/appium/appium/issues/8266#issuecomment-294525667)


dpgraham:
Moved over to driver

PR #228, opened 2017-04-24
Use appium-support zip methods instead of adm-zip

dpgraham:
* `adm-zip` is only used in `commands.pullFolder`
* Added a unit test for `commands.pullFolder`
* Removed `adm-zip` and made the test failed
* Refactored `commands.pullFolder` to use `appium-support` zip methods instead and the tests passed


dpgraham:
https://github.com/appium/appium/issues/7409

dpgraham:
Tagged and published to 1.21.2 (https://www.npmjs.com/package/appium-ios-driver)

PR #231, opened 2017-04-25
Move appium-instruments

imurchie:
First step in https://github.com/appium/appium/issues/7469

Brings `appium-instruments` into `appium-ios-driver`. Unit and functional tests added to the latter's test suite (under driver tests for functional).


imurchie:
You don't want to read over 149 files? :)

PR #233, opened 2017-05-01
Add enablePerformanceLogging cap

imurchie:
There is currently no documented way to turn on iOS performance logging. The code indicates that there is a capability, `loggingPrefs = {performance: true}` which turns it on. Translate this to `enablePerformanceLogging`, which already exists for Android.


imurchie:
Published in `appium-ios-driver@1.23.1`.

PR #235, opened 2017-05-26
Fix import that broke performance logging

imurchie:
Accidentally importing a named export, but only exporting default.

Fixes https://github.com/appium/appium/issues/8546


imurchie:
Published in `appium-ios-driver@1.24.2`.

PR #236, opened 2017-05-29
Do not remove Simulator directory when fullReset is not set

sa-spag:
`resetSimulator` was calling `sim.clean` as `fullResetSimulator` does.


imurchie:
Published in `appium-ios-driver@1.24.3`. Thank you very much for your submission!

PR #238, opened 2017-05-31
Fix Insruments-Without-Delay path

sa-spag:
`getIwdPath` was always fallbacking to `/build/instruments-iwd/iwd`, which is not even existing since `instruments-iwd` directory is not built.


imurchie:
Thanks for this!

imurchie:
Published in `appium-ios-driver@1.24.4`

PR #240, opened 2017-06-20
Add realDeviceLogger capability

imurchie:
Currently we try to use a globally installed `idevicesyslog`, and if that fails we try to use `deviceconsole` which is supposedly installed in Appium's `build` directory. It is not. So that part never works.

Update to allow the user to specify a `realDeviceLogger` capability, which is either:
* nothing (use the system `idevicesyslog`)
* `idevicesyslog` (use the system `idevicesyslog`)
* path to `idevicesyslog`
* `deviceconsole` (use `deviceconsole` in Appium's `build` directory)
* path to `deviceconsole`.

See https://github.com/appium/appium/issues/8698

### DIFF
--- a/lib/device-log/ios-log.js
+++ b/lib/device-log/ios-log.js
@@ -26,45 +26,73 @@ class IOSLog {
   }
 
   async startCaptureRealDevice () {
+    let cmd, args, env;
     if ((this.realDeviceLogger || '').indexOf('idevicesyslog') !== -1) {
       logger.debug('Attempting iOS device log capture via libimobiledevice idevicesyslog');
-      let idevicesyslog;
-      try {
-        if (this.realDeviceLogger.toLowerCase() === 'idevicesyslog') {
-          idevicesyslog = await fs.which('idevicesyslog');
-        } else {
-          idevicesyslog = this.realDeviceLogger;
+      if (this.realDeviceLogger.toLowerCase() === 'idevicesyslog') {
+        cmd = 'idevicesyslog';
+        try {
+          // make sure it is available on the PATH
+          await fs.which('idevicesyslog');
+        } catch (err) {
+          throw new Error(`Unable to find system idevicesyslog: ${err.message}`);
         }
-        logger.debug(`Using idevicesyslog: '${idevicesyslog}'`);
-
-        this.proc = new SubProcess('idevicesyslog', ['-u', this.udid], {env: process.env});
-      } catch (err) {
-        logger.errorAndThrow(`Unable to capture device log using idevicesyslog: ${err.message}`);
+      } else {
+        // make sure the executable exists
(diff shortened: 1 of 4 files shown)


lib/device-log/ios-log.js
```diff
@@ -25,20 +26,46 @@ class IOSLog {
...
+      if (this.realDeviceLogger.toLowerCase() === 'deviceconsole') {
+        deviceconsole = DEVICE_CONSOLE_PATH;
+      } else {
+        deviceconsole = this.realDeviceLogger;
+      }
+      logger.debug(`Using default deviceconsole: '${deviceconsole}'`);
+
+      try {
+        // first make sure the executable is actually there
+        await fs.which(path.resolve(deviceconsole, 'deviceconsole'));
```
jlipps: do we expect the user to know that they should not send in the path of deviceconsole itself but rather its containing directory? i imagine they will probably send in the path of deviceconsole itself...

imurchie: Updated to parse that. Added some unit tests to, because testing.

lib/device-log/ios-log.js
```diff
@@ -25,20 +26,74 @@ class IOSLog {
...
+        }
+        if (stat.isDirectory()) {
+          deviceconsole = this.realDeviceLogger;
+        } else {
+          // make sure they've passed in `deviceconsole` and not something random
+          if (this.realDeviceLogger.lastIndexOf('deviceconsole') !== this.realDeviceLogger.length - 'deviceconsole'.length) {
+            throw new Error(`Unable to parse 'deviceconsole' installation directory from '${this.realDeviceLogger}'`);
+          }
+          // remove the executable, and trailing `/`, to get the install directory
+          deviceconsole = this.realDeviceLogger.substring(0, this.realDeviceLogger.length - 'deviceconsole'.length - 1);
```
jlipps: is there no equivalent of `dirname` in Node's path or fs APIs?

lib/device-log/ios-log.js
```diff
@@ -25,20 +26,74 @@ class IOSLog {
...
+      }
+
+      logger.debug(`Using 'deviceconsole' from '${deviceconsole}'`);
+
+      cmd = 'deviceconsole';
+      args = ['-u', this.udid];
+
+      // set up the environment to be able to run deviceconsole
+      env = _.clone(process.env);
+      env.PATH = `${deviceconsole}:${process.env.PATH}`;
```
jlipps: rather than updating PATH and then calling the command relatively, why not just call the command using its absolute path (which we now know)?

lib/device-log/ios-log.js
imurchie: There is! What fun!

PR #247, opened 2017-08-29
Add fullContextList cap to return context list with url and title

imurchie:
Sometimes it is necessary to know more about the contexts available than just the id, which we currently return.

So, behind a capability `fullContextList`, this PR adds the url and title to the returned list. If this is amendable, I will open a PR to add this to the mobile spec.

Current information returned:
```
[ 'NATIVE_APP', 'WEBVIEW_65515.1' ]
```

With the `fullContextList` capability and this PR:
```
[ { 
    id: 'NATIVE_APP' 
  },
  { 
    id: 'WEBVIEW_69623.1',
    title: 'I am a page title',
    url: 'http://localhost:4994/test/guinea-pig' 
  }
 ]
```

This way if there are multiple webviews the appropriate one can be chosen by the client, rather than needing to cycle through them and query the page to find the correct one.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -25,7 +25,19 @@ commands.getCurrentContext = async function () {
 commands.getContexts = async function () {
   logger.debug('Getting list of available contexts');
   let contexts = await this.getContextsAndViews(false);
-  return contexts.map((context) => context.id.toString());
+
+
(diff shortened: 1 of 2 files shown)


jlipps:
i think this is a worth emendation to the mobile spec

PR #250, opened 2017-10-04
Fix page load logic

imurchie:
Reproduce #197 with latest code.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -317,13 +317,13 @@ extensions.onPageChange = async function (pageChangeNotification) {
       let needsPageLoad = (() => {
         let item = (arr) => {
           return _.filter(arr, (obj) => {
-            return obj.id === this.curContext;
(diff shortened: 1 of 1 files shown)


lib/commands/context.js
```diff
@@ -317,13 +317,13 @@ extensions.onPageChange = async function (pageChangeNotification) {
       let needsPageLoad = (() => {
         let item = (arr) => {
           return _.filter(arr, (obj) => {
```
mykola-mokhnach: wouldn't be https://lodash.com/docs/4.17.4#find more appropriate for such purpose?

imurchie: It might do the same thing, yes. Not sure about "more appropriate".

imurchie: Trying it out...

imurchie:
Published in `appium-ios-driver@1.27.4`.

PR #254, opened 2017-11-02
Make sure we return something useful from getting the date

imurchie:
If we can't parse the date string from `idevicedate`, return something useful (i.e., the output itself).

See https://github.com/appium/appium/issues/7835

### DIFF
--- a/lib/commands/general.js
+++ b/lib/commands/general.js
@@ -27,7 +27,11 @@ commands.getDeviceTime = async function () {
       stdout = stdout.trim();
       try {
         let date = new Date(stdout);
-        return date.toString();
+        let dateStr = date.toString();
+        if (dateStr === 'Invalid Date') {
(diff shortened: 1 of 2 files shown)


lib/commands/general.js
```diff
@@ -27,7 +27,11 @@ commands.getDeviceTime = async function () {
       stdout = stdout.trim();
       try {
         let date = new Date(stdout);
-        return date.toString();
+        let dateStr = date.toString();
+        if (dateStr === 'Invalid Date') {
+          throw new Error('Unable to parse idevicedate output');
```
mykola-mokhnach: should we also print the actual output in the error message?

imurchie: It's returned (and printed) in the `catch` block.

imurchie:
Published in `appium-ios-driver@1.27.5`

PR #261, opened 2018-01-10
Apply w3c timeout

KazuCocoa:
Require adapting https://github.com/appium/appium-base-driver/pull/164 as base driver.
It means we should update the version of the driver in `package.json`

### DIFF
--- a/lib/commands/timeout.js
+++ b/lib/commands/timeout.js
@@ -1,55 +1,27 @@
 import logger from '../logger';
-import { BaseDriver } from 'appium-base-driver';
-import { util } from 'appium-support';
 
 let commands = {}, helpers = {}, extensions = {};
 
-// Arguments will be: [{"protocol":"W3C","implicit":30000}, "1dcfe021-8fc8-49bd-8dac-e986d3091b97", ...]
-// eslint-disable-next-line no-unused-vars
-commands.timeouts = async function (timeoutsObj) {
-  if (timeoutsObj.protocol === BaseDriver.DRIVER_PROTOCOL.W3C) {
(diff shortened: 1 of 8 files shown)


lib/commands/timeout.js
```diff
@@ -1,25 +1,46 @@
...
-      this.setPageLoadTimeout(ms);
-      break;
-    case 'script':
-      this.setAsyncScriptTimeout(ms);
-      break;
-    default:
-      throw new Error(`Invalid timeout '${type}'`);
+// Arguments will be: [{"protocol":"W3C","implicit":30000}, "1dcfe021-8fc8-49bd-8dac-e986d3091b97", ...]
+// eslint-disable-next-line no-unused-vars
+commands.timeouts = async function (timeoutsObj, ...args) {
```
mykola-mokhnach: you can skip ...args declaration

lib/commands/timeout.js
```diff
@@ -1,25 +1,46 @@
...
+
+    switch (type) {
+      case 'command':
+        this.newCommandTimeout(ms);
+        break;
+      case 'implicit':
+        this.implicitWait(ms);
+        break;
+      case 'page load':
+        this.setPageLoadTimeout(this.parseTimeoutArgument(ms));
```
mykola-mokhnach: shall we apply  `parseTimeoutArgument` to all method calls like it was in the previous implementation or to this one only?

KazuCocoa: oops, I forgot to wrap this line...

imurchie:
Thanks for this! Published in `appium-ios-driver@1.31.0`.

PR #263, opened 2018-01-17
Add rect for window

KazuCocoa:
As same as : appium/appium-xcuitest-driver#607

## After

```
       2018-01-20 05:04:30 +0000 Debug: Running system command #4: /Users/kazuaki-matsuo/.nodebrew/node/v8.1.2/bin/node /Users/kazuaki-matsuo/Documents/github/appium/node_modules/appium-ios-driver/build/lib/uiauto/bin/command-proxy-client.js /var/folders/45/rztq_0190zv17z9z32t6msp00000gq/T/instruments_sock 2,{"status":0,"value":{"width":375,"height":667}}...
[debug] [UIAuto] Socket data received (49 bytes)
[debug] [UIAuto] Got result from instruments: {"status":0,"value":{"width":375,"height":667}}
[debug] [MJSONWP] Responding to client with driver.getWindowRect() result: {"width":375,"height":667,"x":0,"y":0}
[HTTP] <-- GET /wd/hub/session/e0db360e-d8bc-49d1-93ff-217022d5bf91/window/rect 200 680 ms - 48 
[BaseDriver] Shutting down because we waited 60 seconds for a command
```


uiauto/lib/mechanic-ext/screen-ext.js
```diff
@@ -59,6 +59,11 @@
       return size;
     }
 
+  , getWindowRect: function () {
```
KazuCocoa: I haven't checked here yet. Just add because this file has `getWindowSize` .

uiauto/lib/mechanic-ext/screen-ext.js
```diff
@@ -59,6 +59,11 @@
...
+      var rect = $.target().rect();
```
mykola-mokhnach: this could be returned immediately

PR #265, opened 2018-02-13
Pass in xcode version to ios logger, rather than computing again

imurchie:
We already have the xcode version at the point we instantiate the logger, so just pass it in, since this is not a trivial operation.

Related (tangentially) to the work of https://github.com/appium/appium/issues/10188

### DIFF
--- a/lib/device-log/ios-log.js
+++ b/lib/device-log/ios-log.js
@@ -2,7 +2,6 @@ import path from 'path';
 import _ from 'lodash';
 import logger from './logger';
 import { fs, mkdirp } from 'appium-support';
-import xcode from 'appium-xcode';
 import { SubProcess, exec } from 'teen_process';
 
 const START_TIMEOUT = 10000;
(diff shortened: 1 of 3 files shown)


lib/device-log/ios-log.js
```diff
@@ -17,6 +16,7 @@ class IOSLog {
     this.udid = opts.udid;
```
mykola-mokhnach: Where the logger is created?

imurchie: In this driver, it is here: https://github.com/appium/appium-ios-driver/blob/246af060fc491af7762142269b84a9ce3638748a/lib/commands/logging.js#L54-L60

In `appium-xcuitest-driver` it is here: https://github.com/appium/appium-xcuitest-driver/blob/master/lib/commands/log.js#L22-L27

This will necessitate a major bump when published, so the xcuitest driver can be updated.

mykola-mokhnach: So you mean the `xcodeVersion` property has been already set for iOS driver

imurchie: Yes. It gets set in session creation.

PR #266, opened 2018-02-19
Cache idevicesyslog processes per UDID

dpgraham:
(connected to https://github.com/appium/appium/issues/7838)

### DIFF
--- a/lib/device-log/ios-log.js
+++ b/lib/device-log/ios-log.js
@@ -282,10 +282,12 @@ class IOSLog {
    */
   getIDeviceSysLogProcess (cmd, args, env) {
     if (!IOSLog.cachedIDeviceSysLogs[this.subprocessId]) {
-      IOSLog.cachedIDeviceSysLogs[this.subprocessId] = {
+      let {proc} = IOSLog.cachedIDeviceSysLogs[this.subprocessId] = {
         proc: new SubProcess(cmd, args, {env}),
         count: 1,
       };
+      // If the process dies, remove it from the cache
+      proc.on('exit', () => delete IOSLog.cachedIDeviceSysLogs[this.subprocessId]);
     } else {
       IOSLog.cachedIDeviceSysLogs[this.subprocessId].count++;
     }
--- a/lib/device-log/ios-log.js
+++ b/lib/device-log/ios-log.js
@@ -88,15 +88,17 @@ class IOSLog {
       logger.errorAndThrow(`Unable to capture device log. Unknown 'realDeviceLogger': '${this.realDeviceLogger}'`);
     }
 
+    this.subprocessId = `${cmd}_${this.udid}`;
+
     logger.debug(`Starting iOS device log capture with: '${cmd}'`);
 
     try {
       // cleanup existing listeners if the previous session has not been terminated properly
-      await exec('pkill', ['-xf', [cmd, ...args].join(' ')]);
+      await this.killExistingSysLogProcesses(cmd, args);
     } catch (e) {}
 
     // just use one instance of 'idevicesyslog' per udid
-    this.proc = IOSLog._getIDeviceSysLogProcess(cmd, args, env, this.udid);
+    this.proc = this.getIDeviceSysLogProcess(cmd, args, env);
 
     await this.finishStartingLogCapture();
   }
@@ -201,8 +203,23 @@ class IOSLog {
   }
 
   async stopCapture () {
-    logger.debug('Stopping iOS log capture');
-    if (this.proc && this.proc.isRunning) {
+    if (this.proc) {
+      if (this.udid) {
+        // If no other UDID's are using 'idevicesyslog' kill it
+        if (--IOSLog.cachedIDeviceSysLogs[this.subprocessId].count === 0) {
+          await this.killLogSubProcess();
+          delete IOSLog.cachedIDeviceSysLogs[this.subprocessId];
+        }
+      } else {
+        await this.killLogSubProcess();
+      }
+    }
+    this.proc = null;
+  }
+
+  async killLogSubProcess () {
+    if (this.proc.isRunning) {
+      logger.debug('Stopping iOS log capture');
       try {
         await this.proc.stop('SIGTERM', 1000);
       } catch (e) {
@@ -210,8 +227,6 @@ class IOSLog {
         await this.proc.stop('SIGKILL');
       }
     }
-    this.proc = null;
-    IOSLog.cachedIDeviceSysLogs[this.udid] = null;
   }
 
   onOutput (prefix = '') {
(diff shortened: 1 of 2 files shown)


lib/device-log/ios-log.js
```diff
@@ -249,7 +252,25 @@ class IOSLog {
...
     return this.logs;
   }
+
+  /**
+   * Get 'idevicesyslog' subprocess for a given udid.
+   * 
+   * This function caches and re-uses instances so that there's
+   * just one instance per device (cache by UDID)
+   */
+  static async _getIDeviceSysLogProcess (cmd, args, env, udid) {
```
jlipps: why is this an async fn?

lib/device-log/ios-log.js
```diff
@@ -249,7 +252,25 @@ class IOSLog {
...
+    if (!IOSLog.cachedIDeviceSysLogs[cmd]) {
```
jlipps: doesn't really matter, but we could get away with a 1-level object doing something like 

```js
IOSLog.cachedIDeviceSysLogs[`${cmd}_${udid}`]
```

lib/device-log/ios-log.js
dpgraham: Just a habit. I thought we had a rule to catch this.

mykola-mokhnach:
Now it is not the case anymore, because we kill obsolete processes anyway (see the line above:

>        // cleanup existing listeners if the previous session has not been terminated properly
>        await exec('pkill', ['-xf', [cmd, ...args].join(' ')]);

)

lib/device-log/ios-log.js
```diff
@@ -249,7 +252,25 @@ class IOSLog {
...
+      IOSLog.cachedIDeviceSysLogs[cmd] = {};
+    }
+    if (!IOSLog.cachedIDeviceSysLogs[cmd][udid]) {
+      IOSLog.cachedIDeviceSysLogs[cmd][udid] = new SubProcess(cmd, args, {env});
+    }
+    return IOSLog.cachedIDeviceSysLogs[cmd][udid];
```
mykola-mokhnach: here we don't actually check if the process is running even if it is in the cache. In general, it should be removed from the cache as soon as it is terminated.

lib/device-log/ios-log.js
```diff
@@ -209,6 +211,7 @@ class IOSLog {
       }
     }
     this.proc = null;
+    IOSLog.cachedIDeviceSysLogs[this.udid] = null;
```
mykola-mokhnach: this object structure is in conflict with `_getIDeviceSysLogProcess`m since there `cmd` is used as the key

dpgraham: Good catch.

dpgraham:
@mykola-mokhnach The problem people were having was that when they called `launchApp` more than once on the same device, each time it would spawn a new instance of `idevicesyslog`. This PR verifies that there are no redundant instances of `idevicesyslog`.

lib/device-log/ios-log.js
dpgraham: My first implementation only used the `udid`. I added the `cmd` key in-case there were different subprocesses running off the same `udid`. I can change it to only use `udid` if you don't think there's any issue with that.

lib/device-log/ios-log.js
dpgraham: Makes sense. I should bind a method to the `exit` event that deletes the reference to the SubProcess?

lib/device-log/ios-log.js
imurchie: We _will_ have a rule. It requires updating eslint on all packages. So we do not at the moment. :|

lib/device-log/ios-log.js
```diff
@@ -249,7 +252,26 @@ class IOSLog {
...
   }
+
+  /**
+   * Get 'idevicesyslog' subprocess for a given udid.
+   * 
+   * This function caches and re-uses instances so that there's
+   * just one instance per device (cache by UDID)
+   */
+  static _getIDeviceSysLogProcess (cmd, args, env, udid) {
+    if (!IOSLog.cachedIDeviceSysLogs[udid]) {
```
mykola-mokhnach: perhaps here you should also check if the process is currently running

dpgraham:
I just pushed fixes.

lib/device-log/ios-log.js
dpgraham: Good thing you spotted this @jlipps. I had assigned `proc` to the promise returned by `_getIDeviceSysLogProcess` instead of the actual `SubProcess`.

dpgraham:
Hey @jlipps, @imurchie and @mykola-mokhnach,

I had to go back to refactor this quite a bit due to some issues I noticed. 

The main thing is that I had to change it so that it counts how many instances of `idevicesyslog` are open so that and when `stopCapture` is called, it only kills the process if the count is 0.

lib/device-log/ios-log.js
```diff
@@ -199,16 +203,30 @@ class IOSLog {
   }
 
   async stopCapture () {
-    logger.debug('Stopping iOS log capture');
-    if (this.proc && this.proc.isRunning) {
+    if (this.proc) {
+      if (this.udid) {
```
mykola-mokhnach: can it be that udid is not set?

lib/device-log/ios-log.js
```diff
@@ -249,7 +267,33 @@ class IOSLog {
...
+   * just one instance per device (cache by UDID)
+   */
+  getIDeviceSysLogProcess (cmd, args, env) {
+    if (!IOSLog.cachedIDeviceSysLogs[this.subprocessId]) {
+      IOSLog.cachedIDeviceSysLogs[this.subprocessId] = {
+        proc: new SubProcess(cmd, args, {env}),
+        count: 1,
+      };
+    } else {
+      IOSLog.cachedIDeviceSysLogs[this.subprocessId].count++;
```
mykola-mokhnach: this counter should not be increased if the process exists in the cache, but is not running, although it is not checked

lib/device-log/ios-log.js
dpgraham: Checking `this.udid` checks if it's a real device (which `idevicesyslog` runs on)

lib/device-log/ios-log.js
dpgraham: @mykola-mokhnach Is there ever a case where the process wouldn't be running?

mykola-mokhnach: yep, the process might die if the device has been unexpectedly disconnected or it has been killed manually or just because of a bug

dpgraham: Maybe something like

```javascript
proc.on('exit', () => delete IOSLog.cachedIDeviceSysLogs[this.subprocessId]);
```

be a good approach?

dpgraham: @mykola-mokhnach I added a fix for this.

mykola-mokhnach:
it looks like one unit test has been broken

dpgraham:
@mykola-mokhnach Should be fixed. Needed to stub `fs.which` to return `true`.

mykola-mokhnach:
Ok, lets see how it works

lib/device-log/ios-log.js
```diff
@@ -199,16 +203,30 @@ class IOSLog {
   }
 
   async stopCapture () {
-    logger.debug('Stopping iOS log capture');
-    if (this.proc && this.proc.isRunning) {
+    if (this.proc) {
+      if (this.udid) {
+        // If no other UDID's are using 'idevicesyslog' kill it
+        if (--IOSLog.cachedIDeviceSysLogs[this.subprocessId].count === 0) {
```
imurchie: What is this post-decrement doing? Seems like a side-effect waiting to cause a bug. If this `count` needs to be decremented, I'd rather it happen explicitly, rather than in the conditional.

dpgraham:
Fix pushed.

PR #268, opened 2018-02-23
Update for base driver logging

imurchie:
Use logging functions from https://github.com/appium/appium-base-driver/pull/191. Supersedes https://github.com/appium/appium-ios-driver/pull/267

### DIFF
--- a/test/e2e/driver/logging-e2e-specs.js
+++ b/test/e2e/driver/logging-e2e-specs.js
@@ -4,7 +4,6 @@ import env from '../helpers/env';
 import { rootDir } from '../../../lib/utils';
 import path from 'path';
 import { IosDriver } from '../../..';
-import { SUPPORTED_LOG_TYPES } from '../../../lib/commands/logging.js';
 import chai from 'chai';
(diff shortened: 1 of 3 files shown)


test/e2e/driver/logging-e2e-specs.js
```diff
@@ -25,7 +24,7 @@ describe('commands - logging', function () {
   describe('getLogTypes', function () {
     it('should get the list of available logs', async function () {
       driver.getLogTypes.should.an.instanceof(Function);
-      (await driver.getLogTypes()).should.eql(_.keys(SUPPORTED_LOG_TYPES));
+      (await driver.getLogTypes()).should.eql(_.keys(driver.supportedLogTypes));
```
mykola-mokhnach: is keys order always preserved?

imurchie: I think it is, though it's not guaranteed. But this test is not new and hasn't failed before.

But I could change it to how I did it in the android driver (i.e., an exclusive or of the two ought to be empty).

PR #269, opened 2018-03-12
Abstract out remote debugger creation

imurchie:
So subclasses can instantiate it differently. The function is `async` so subclasses can do async-y things, even if it is not necessary here.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -169,6 +169,15 @@ extensions.getContextsAndViews = async function (useUrl = true) {
   return ctxs;
 };
 
+extensions.getNewRemoteDebugger = async function () {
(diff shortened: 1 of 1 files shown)


lib/commands/context.js
```diff
@@ -169,6 +169,15 @@ extensions.getContextsAndViews = async function (useUrl = true) {
   return ctxs;
 };
 
+extensions.getNewRemoteDebugger = async function () {
```
mykola-mokhnach: should it be async?

imurchie: Please see PR description (https://github.com/appium/appium-ios-driver/pull/269#issue-174448041).

PR #273, opened 2018-05-03
Update dependencies

imurchie:


package.json
```diff
@@ -29,7 +29,7 @@
   },
   "dependencies": {
     "appium-base-driver": "^2.23.0",
-    "appium-ios-simulator": "^1.10.3",
+    "appium-ios-simulator": "^2.12.0",
```
mykola-mokhnach: what was changed in v2?

imurchie: The breaking change was a refactoring of the arguments to `run` (https://github.com/appium/appium-ios-simulator/commit/e7fd0faa9811d9f47a6a630bc505b728b4b60e50).

mykola-mokhnach: so this does not affect ios driver?

imurchie: Only in so far as it will be able to partake in improvements that are made. Nothing is changed to the functioning.

PR #295, opened 2018-07-26
Clear frames when a new page loads.

tonyhignett:
- Otherwise we still try to run atoms in the frame, which fails.
- Should fix appium/appium#11102.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -330,6 +330,9 @@ extensions.onPageChange = async function (pageChangeNotification) {
         await this.remote.pageLoad();
       }
 
+      logger.debug("Clearing any frames");
+      this.curWebFrames = [];
+
       logger.debug('New page listing is same as old, doing nothing');
     }
   }


tonyhignett:
I've tested that this resolves the issue, but I don't understand the code well so I'm not sure it's the right fix. In particular, 
- I can't manage to exercise any of the branches of `onPageChange` except the one I've changed. I don't know how to get a "new page" or a "closed window".
- this will clear frames when you navigate to an anchor in the same page, which is probably not what you expect. But the block headed "If a window navigates to an anchor" doesn't detect that, although the comment sounds like it does. Is that a bug?

lib/commands/context.js
```diff
@@ -330,6 +330,9 @@ extensions.onPageChange = async function (pageChangeNotification) {
         await this.remote.pageLoad();
       }
 
+      logger.debug("Clearing any frames");
+      this.curWebFrames = [];
```
jlipps: seems reasonable to clear the frames if the page changes or reloads. should we not put this into the `if (needsPageLoad)` conditional above, though? otherwise we will clear the frames even when we don't think anything has happened

tonyhignett: The trouble is that I can't get it to go into that branch. And looking at the code closer I don't see how it could. `needsPageLoad` is "is `this.curContext` in exactly one of `this.contexts` and `pageArray`?", but we have synced those lists at line 277.

tonyhignett: > it could theoretically happen in between a user selecting a frame and trying to do something with it.

In a way that's exactly the problem. The user selects a frame and then tries to run some javascript in it, but in the meantime a new page has unexpectedly loaded under their feet, so the frame no longer exists.

jlipps: I see. So better for Appium not to think the frame exists and to return a NoSuchFrame error so the user can retry. Alright, I'm cool with this. I still want @imurchie's review though.

imurchie: I'm exploring this this morning.

imurchie:
So... this will not work, unfortunately. The Web Inspector sends too many events for page change, when there is no change. As far as I can see there is no way to distinguish them.

imurchie:
@tonyhignett Can you try the latest Appium beta (with `appium-ios-driver@2.6.0`)?

tonyhignett:
@imurchie It's working nicely in a manual test. Will take longer to be sure it's working in our production environment, but I'm happy to withdraw this in its favour.

Thanks for the fix!

PR #297, opened 2018-07-30
add remoteDebugProxy to descaps

yfangsl:
All further info regarding this pull request can be found at appium/appium#10960

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -174,7 +174,8 @@ extensions.getNewRemoteDebugger = async function () {
     bundleId: this.opts.bundleId,
     useNewSafari: this.useNewSafari(),
     pageLoadMs: this.pageLoadMs,
-    platformVersion: this.opts.platformVersion
+    platformVersion: this.opts.platformVersion,
+    remoteDebugProxy: this.opts.remoteDebugProxy
   });
 };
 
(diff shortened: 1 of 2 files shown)


imurchie:
Does this not need to be passed into the RemoteDebugger? 

https://github.com/appium/appium-ios-driver/blob/master/lib/commands/context.js#L172-L179

imurchie:
`appium-xcuitest-driver` uses parts of `appium-ios-driver`, and extends its functionality. If the cap is in `appium-ios-driver` then it will be accepted by both, as `appium-xcuitest-driver` pulls all the ios driver's caps and then adds some.

But both drivers implement a version of `getNewRemoteDebugger` because they need different parameters (like the socket path, which the ios driver knows nothing about, and should not know anything about):
* `appium-ios-driver`: https://github.com/appium/appium-ios-driver/blob/master/lib/commands/context.js#L172-L179
* `appium-xcuitest-driver`: https://github.com/appium/appium-xcuitest-driver/blob/master/lib/commands/context.js#L63-L73

ThomasFettSauce:
Thanks for the clarification!
Then I agree that we pass the capability also here, so that we cover both cases:
https://github.com/appium/appium-ios-driver/blob/master/lib/commands/context.js#L172-L179

yfangsl:
Got it, will add it in today

PR #300, opened 2018-08-03
Remove frames when RD alerts

imurchie:
Alternative to #295

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -224,6 +224,10 @@ extensions.listWebFrames = async function (useUrl = true) {
     }
     pageArray = await this.remote.selectApp(currentUrl, this.opts.webviewConnectRetries, this.opts.ignoreAboutBlankUrl);
     this.remote.on(RemoteDebugger.EVENT_PAGE_CHANGE, this.onPageChange.bind(this));
+    this.remote.on(RemoteDebugger.EVENT_FRAMES_DETACHED, () => {
+      logger.debug(`Clearing ${this.curWebFrames.length} frames: ${this.curWebFrames.join(', ')}`);
+      this.curWebFrames = [];
+    });
 
     let tryClosingAlert = async () => {
       let didDismiss = await this.closeAlertBeforeTest();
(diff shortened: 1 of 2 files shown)


lib/commands/context.js
```diff
@@ -224,6 +224,10 @@ extensions.listWebFrames = async function (useUrl = true) {
     }
     pageArray = await this.remote.selectApp(currentUrl, this.opts.webviewConnectRetries, this.opts.ignoreAboutBlankUrl);
     this.remote.on(RemoteDebugger.EVENT_PAGE_CHANGE, this.onPageChange.bind(this));
+    this.remote.on(RemoteDebugger.EVENT_FRAMES_DETACHED, () => {
+      logger.debug(`Clearing ${this.curWebFrames.length} frames: ${this.curWebFrames.join(', ')}`);
```
mykola-mokhnach: I think there is only sense to log is if `this.curWebFrames` is not empty

lib/commands/context.js
```diff
@@ -285,53 +291,55 @@ extensions.onPageChange = async function (pageChangeNotification) {
...
+    this.curContext = `${appIdKey}.${keyId}`;
+    newPage = keyId;
+  } else {
+    // at this point, there are no new pages, and the current page still exists
+    logger.debug('Checking if page needs to load');
+    // If a window navigates to an anchor it doesn't always fire a page
+    // callback event. Let's check if we wound up in such a situation.
+    const needsPageLoad = (() => {
+      // need to map the page ids to context ids
+      const contextArray = _.map(pageArray, (arr) => `${appIdKey}.${arr.id}`);
```
mykola-mokhnach: parameter name `arr` is confusing

lib/commands/context.js
```diff
@@ -285,53 +291,55 @@ extensions.onPageChange = async function (pageChangeNotification) {
...
+      // check if the current context exists in both our recorded contexts,
+      // and the page array
+      return !_.isEqual(_.find(this.contexts, this.curContext), _.find(contextArray, this.curContext));
```
mykola-mokhnach: is it possible that both _.find functions return `undefined` ?

PR #325, opened 2018-10-10
Clean up window handle code

imurchie:
We want to not filter by url, or else we don't actually get all the window handles (see https://github.com/appium/appium/issues/11484).

Further, `this.massagePage` does not exist, so mapping through it does not do anything at all. This was a holdover from Appium 1.4 (https://github.com/appium/appium/blob/1.4/lib/devices/ios/ios-controller.js#L1717-L1720).

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -26,7 +26,6 @@ commands.getContexts = async function () {
   logger.debug('Getting list of available contexts');
   let contexts = await this.getContextsAndViews(false);
 
-
   let mapFn = (context) => context.id.toString();
   if (this.opts.fullContextList) {
     mapFn = (context) => {
@@ -109,9 +108,8 @@ commands.getWindowHandles = async function () {
     throw new errors.NotImplementedError();
   }
 
-  let pageArray = await this.listWebFrames();
-  this.windowHandleCache = _.map(pageArray, this.massagePage);
-  let idArray = _.map(this.windowHandleCache, 'id');
+  this.windowHandleCache = await this.listWebFrames(false);
+  const idArray = _.map(this.windowHandleCache, 'id');
   // since we use this.contexts to manage selecting debugger pages, make
   // sure it gets populated even if someone did not use the
   // getContexts method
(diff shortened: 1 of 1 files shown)


lib/commands/context.js
```diff
@@ -109,9 +108,8 @@ commands.getWindowHandles = async function () {
     throw new errors.NotImplementedError();
   }
 
-  let pageArray = await this.listWebFrames();
-  this.windowHandleCache = _.map(pageArray, this.massagePage);
```
mykola-mokhnach: perhaps it should be `this.mEssagePage`?

imurchie: I don't understand.

lib/commands/context.js
```diff
@@ -109,9 +108,8 @@ commands.getWindowHandles = async function () {
...
-  let idArray = _.map(this.windowHandleCache, 'id');
+  this.windowHandleCache = await this.listWebFrames(false);
+  const idArray = _.map(this.windowHandleCache, 'id');
```
mykola-mokhnach: I cannot understand what this _.map is doing

imurchie: It's pulling the `id` property out of the array.

imurchie: It uses the `_.property` shorthand. https://lodash.com/docs/4.17.10#property

lib/commands/context.js
mykola-mokhnach: I mean the property name is misspelled

lib/commands/context.js
mykola-mokhnach: it would be useful to have some docs on how `listWebFrames` result looks like

mykola-mokhnach: I'd also rename this variable to `frameIds` or something like that

lib/commands/context.js
imurchie: Oh. No. `massagePane` used to exist. It was never ported over when we rewrote.

PR #337, opened 2018-12-10
Use W3C and MJSONWP element keys

dpgraham:
When converting elements, it should include MJSONWP and ELEMENT

### DIFF
--- a/lib/commands/web.js
+++ b/lib/commands/web.js
@@ -387,13 +387,19 @@ helpers.useAtomsElement = function (el) {
 helpers.convertElementsForAtoms = function (args = []) {
   let newArgs = [];
   for (let arg of args) {
-    if (!_.isNull(arg) && !_.isUndefined(arg.ELEMENT)) {
-      let atomsElement = this.getAtomsElement(arg.ELEMENT);
-      if (atomsElement === null) {
-        throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+    let hasElementKey = false;
+    for (let elementKey of ['ELEMENT', 'element-6066-11e4-a52e-4f735466cecf']) {
+      if (!_.isNull(arg) && !_.isUndefined(arg[elementKey])) {
+        hasElementKey = true;
+        let atomsElement = this.getAtomsElement(arg[elementKey]);
+        if (atomsElement === null) {
+          throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+        }
+        newArgs.push(atomsElement);
       }
-      newArgs.push(atomsElement);
-    } else {
+    }
+
+    if (!hasElementKey) {
       newArgs.push(arg);
     }
   }
--- a/lib/commands/web.js
+++ b/lib/commands/web.js
@@ -370,7 +370,7 @@ helpers.getAtomsElement = function (wdId) {
   if (_.isUndefined(atomsId)) {
     return null;
   }
-  return {ELEMENT: atomsId, [W3C_WEB_ELEMENT_IDENTIFIER]: atomsId};
+  return {ELEMENT: atomsId};
 };
 
 helpers.useAtomsElement = function (el) {
@@ -440,7 +440,7 @@ helpers.parseElementResponse = function (element) {
   let objId = helpers.getElementId(element);
   let clientId = (ELEMENT_OFFSET + this.webElementIds.length).toString();
   this.webElementIds.push(objId);
-  return {ELEMENT: clientId, [W3C_WEB_ELEMENT_IDENTIFIER]: clientId};
+  return {ELEMENT: clientId};
 };
 
 helpers.getElementId = function (element) {
--- a/lib/commands/web.js
+++ b/lib/commands/web.js
@@ -9,6 +9,7 @@ import logger from '../logger';
 let commands = {}, helpers = {}, extensions = {};
 
 const ELEMENT_OFFSET = 5000;
+const { W3C_WEB_ELEMENT_IDENTIFIER } = util;
 
 commands.setFrame = async function (frame) {
   if (!this.isWebContext()) {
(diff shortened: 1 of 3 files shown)


lib/commands/web.js
```diff
@@ -387,13 +387,19 @@ helpers.useAtomsElement = function (el) {
...
+        if (atomsElement === null) {
+          throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+        }
+        newArgs.push(atomsElement);
       }
-      newArgs.push(atomsElement);
-    } else {
+    }
+
+    if (!hasElementKey) {
```
mykola-mokhnach: why do we need hasElementKey flag if we can simply put it into `else`?

lib/commands/web.js
```diff
@@ -387,13 +387,19 @@ helpers.useAtomsElement = function (el) {
...
-        throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+    let hasElementKey = false;
+    for (let elementKey of ['ELEMENT', 'element-6066-11e4-a52e-4f735466cecf']) {
+      if (!_.isNull(arg) && !_.isUndefined(arg[elementKey])) {
+        hasElementKey = true;
+        let atomsElement = this.getAtomsElement(arg[elementKey]);
+        if (atomsElement === null) {
+          throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+        }
+        newArgs.push(atomsElement);
```
mykola-mokhnach: is it possible that both keys are present and the element would be added twice?

lib/commands/web.js
mykola-mokhnach: https://www.w3schools.com/js/js_break.asp

lib/commands/web.js
dpgraham: I forgot to take out `newArgs.push(atomsElement)` from the for-loop.

lib/commands/web.js
```diff
@@ -387,13 +387,19 @@ helpers.useAtomsElement = function (el) {
 helpers.convertElementsForAtoms = function (args = []) {
   let newArgs = [];
   for (let arg of args) {
-    if (!_.isNull(arg) && !_.isUndefined(arg.ELEMENT)) {
-      let atomsElement = this.getAtomsElement(arg.ELEMENT);
-      if (atomsElement === null) {
-        throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg.ELEMENT}'`);
+    let hasElementKey = false;
+    for (let elementKey of ['ELEMENT', 'element-6066-11e4-a52e-4f735466cecf']) {
```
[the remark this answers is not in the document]
dpgraham: Appium Support exposes it. I'll use that.

lib/commands/web.js
dpgraham: It shouldn't have been in the for-loop.

dpgraham:
Fixes pushed.

lib/commands/web.js
```diff
@@ -387,13 +388,18 @@ helpers.useAtomsElement = function (el) {
...
+        let atomsElement = this.getAtomsElement(arg[elementKey]);
+        if (atomsElement === null) {
+          throw new errors.UnknownError(`Error converting element ID for using in WD atoms: '${arg[elementKey]}'`);
+        }
       }
-      newArgs.push(atomsElement);
-    } else {
+    }
+
+    if (!hasElementKey) {
```
jlipps: i still don't think this will work. as it stands, the code will not add args to the args list which are _not_ elements. we still want everything that was in `args` to be in `newArgs`, it's just that elements should be converted along the way.

this is also doing nothing with `atomsElement`, which is the thing we want to add to `newArgs`

i think a better solution would involve a `break` in the inner loop to ensure we only push one converted element.

dpgraham: I'll write a test for this.

lib/commands/web.js
```diff
@@ -433,12 +437,20 @@ helpers.parseExecuteResponse = function (res) {
...
 
 helpers.parseElementResponse = function (element) {
-  let objId = element.ELEMENT;
+  let objId = helpers.getElementId(element);
   let clientId = (ELEMENT_OFFSET + this.webElementIds.length).toString();
   this.webElementIds.push(objId);
   return {ELEMENT: clientId};
 };
 
+helpers.getElementId = function (element) {
```
jlipps: this looks good but is it the same as unwrapElement?

dpgraham: It is yeah. I'll update that.

lib/commands/web.js
```diff
@@ -433,12 +437,20 @@ helpers.parseExecuteResponse = function (res) {
...
+  return element.ELEMENT || element[W3C_WEB_ELEMENT_IDENTIFIER];
+};
+
+helpers.hasElementId = function (element) {
+  return util.hasValue(element) && (util.hasValue(element.ELEMENT) || util.hasValue(element[W3C_WEB_ELEMENT_IDENTIFIER]));
```
mykola-mokhnach: and this can be also changed to `return util.unwrapElement(element) !== element`

test/e2e/safari/webview/web-specs.js
```diff
@@ -0,0 +1,29 @@
...
+      let primitives = ['hello', 1, true];
+      driver.convertElementsForAtoms(primitives).should.eql(primitives);
+    });
+    it('should parse W3C and MJSONWP element ids', function () {
+      for (const elType of ['ELEMENT', util.W3C_WEB_ELEMENT_IDENTIFIER]) {
+        driver.webElementIds = [123];
+        const args = [{
+          [elType]: 5000
+        }];
+        driver.convertElementsForAtoms(args).should.eql([{ELEMENT: 123, [util.W3C_WEB_ELEMENT_IDENTIFIER]: 123}]);
```
imurchie: This line is breaking the build. I'm not sure what the intention is in `convertElementsForAtoms`, so I cannot fix it.

dpgraham: I'll look at it today

dpgraham: I'll look at it today

jlipps: the point of `convertElementForAtoms` is to unwrap an element, subtract 5000 from its numeric id, and rewrap it.

the reason is that the selenium atoms maintain their own element id list, which would conflict with the driver's element id cache. so we add 5000 to the selenium atom element id before returning it to the client. that means we also have to subtract it before reusing it in the atoms context.

jlipps: this test should probably be fixed by making the arg `5123` instead?

imurchie: I think the test should not have the W3C part in the expectation, since the atoms don't expect it. But that's the part where I don't know what the intention of the test or the changes to the function are.

dpgraham: I put in a fix for that. The atom doesn't include the W3C element format, and I put in a fix to reflect that.

PR #341, opened 2019-01-09
Move authorize-ios into package

imurchie:
The `authorize-ios` utility is only used for `appium-ios-driver`, so there is no need to have it be separate. 

The final vestiges of dearchitecturing.


imurchie:
Probably worth doing, though it's not technically a breaking change.

PR #362, opened 2019-03-14
Replace outdated dependency with a new-born

MrDatastorage:


mykola-mokhnach:
I don't observe many fixes has been added into the new fork. Isn't it possible to continue working with the original lib?

MrDatastorage:
> I don't observe many fixes has been added into the new fork. Isn't it possible to continue working with the original lib?

@mykola-mokhnach, we going to maintenance the xmldom. Only the old xmldom is never been updated in 2 years. A lot of contributors love to keep maintaining. So we great the reborn.

PR #368, opened 2019-04-01
pass platformversion and issafari to WebKitRemoteDebugger in real device

KazuCocoa:
When I investigated https://github.com/appium/appium/issues/12398#issuecomment-478270634 , I found this.

In https://github.com/appium/appium-remote-debugger/pull/122/files#diff-79c9727f17e7f83c1c33f14bbe7a55e2R23 , the platfotmVersion should be `12.1` for example.
But in real device, `WebKitRemoteDebugger` does not pass the value to the class. So, `opts.platformVersion` becomes `undefined`. As the result, https://github.com/appium/appium-remote-debugger/pull/122/files#diff-17ff41d4312838bfdfc249e22bb720f8R43 never be `true`.

After this change, I ensured https://github.com/appium/appium/issues/12398#issuecomment-478270634 did not happen in my environment.

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -203,6 +203,8 @@ extensions.listWebFrames = async function listWebFrames (useUrl = true) {
       this.remote = new WebKitRemoteDebugger({
         port: this.opts.webkitDebugProxyPort,
         webkitResponseTimeout: this.opts.webkitResponseTimeout,
+        platformVersion: this.opts.platformVersion,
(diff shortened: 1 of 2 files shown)


KazuCocoa:
ah, I see about the next step 👍 
Will bump package.json after publishing new https://github.com/appium/appium-remote-debugger

imurchie:
`appium-remote-debugger@4.1.0` is out.

PR #373, opened 2019-04-26
Mock xcrun calls

imurchie:
Three unit tests still actually made system calls to `xcrun`. Mock those.

@mykola-mokhnach Can you run the existing utils unit tests (`npm run build && npx mocha -R spec build/test/unit/utils-specs.js`) before trying this patch? If the same three `TypeError` errors come up, then that kind of confirms this is the place. Then with this patch those should be gone.

That would also confirm that the problem is somewhere in the system call, which sucks. I would suggest at that point going into `teen_process` and trying to `exec` the `xcrun --sdk iphonesimulator --show-sdk-version` and see where the error comes up.

I wish I could reproduce this so I could investigate. :(

### DIFF
--- a/test/unit/utils-specs.js
+++ b/test/unit/utils-specs.js
@@ -1,26 +1,34 @@
-// transpile:mocha
-
 import * as utils from '../../lib/utils';
 import chai from 'chai';
 import chaiAsPromised from 'chai-as-promised';
+import { withMocks } from 'appium-test-support';
+import xcode from 'appium-xcode';
+
 
 chai.should();
 chai.use(chaiAsPromised);
 
-describe('prepareIosOpts', function () {
+describe('prepareIosOpts', withMocks({xcode}, (mocks) => {
+  beforeEach(function () {
+    mocks.xcode.expects('getMaxIOSSDK')
+      .once().returns('9.3');
+  });
+  afterEach(function () {
+    mocks.verify();
+  });
   it('should use instruments without delay by default', function () {
-    let opts = {};
+    const opts = {};
     utils.prepareIosOpts(opts);
     opts.withoutDelay.should.be.true;
   });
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
```
npm run build && npx mocha -R spec build/test/unit/utils-specs.js

> appium-ios-driver@4.2.2 build /Users/elf/code/appium-ios-driver
> gulp transpile

[07:46:20] Using gulpfile ~/code/appium-ios-driver/gulpfile.js
[07:46:20] Starting 'transpile'...
[07:46:24] Finished 'transpile' after 4.01 s


  prepareIosOpts
    ✓ should use instruments without delay by default
    ✓ should use instruments without delay if explicitly not using native instruments
    ✓ should not use instruments without delay if using native intruments


  3 passing (22ms)
```

There are fine

mykola-mokhnach:
Also I cannot imagine how undefined could be returned from teen_process, because it can ONLY resolve to a dictionary: https://github.com/appium/node-teen_process/blob/30ab25a454b19507a3050fa9dd6613d6edb6470b/lib/exec.js#L92

test/unit/utils-specs.js
```diff
@@ -1,26 +1,34 @@
...
   it('should use instruments without delay if explicitly not using native instruments', function () {
-    let opts = {nativeInstrumentsLib: false};
+    const opts = {nativeInstrumentsLib: false};
     utils.prepareIosOpts(opts);
     opts.withoutDelay.should.be.true;
   });
   it('should not use instruments without delay if using native intruments', function () {
-    let opts = {nativeInstrumentsLib: true};
+    const opts = {nativeInstrumentsLib: true};
     utils.prepareIosOpts(opts);
```
mykola-mokhnach: this is an async call, btw

PR #378, opened 2019-05-24
Add safariShowFullResponse cap

imurchie:
For logging verbosity reasons we truncate the data the Web Inspector returns, but it is sometimes useful to see it, in full. This cap gives access to that information.


lib/commands/context.js
```diff
@@ -179,6 +179,9 @@ extensions.getNewRemoteDebugger = async function getNewRemoteDebugger () { // es
     garbageCollectOnExecute: util.hasValue(this.opts.safariGarbageCollect)
       ? !!this.opts.safariGarbageCollect
       : true,
+    logFullResponse: util.hasValue(this.opts.safariShowFullResponse)
```
mykola-mokhnach: isn't this the same as `!!this.opts.safariShowFullResponse` ?

PR #382, opened 2019-06-14
Properly return whether settings were set

imurchie:
Reduce the chance of having to restart a simulator, when there is not actually a need to.


lib/settings.js
```diff
@@ -139,24 +142,24 @@ async function setSafariPrefs (sim, opts = {}) {
...
   if (_.has(opts, 'safariOpenLinksInBackground')) {
-    let val = opts.safariOpenLinksInBackground ? 1 : 0;
+    const val = opts.safariOpenLinksInBackground ? 1 : 0;
     logger.debug(`Setting opening links in background to '${!!val}'`);
     safariSettings.OpenLinksInBackground = val;
   }
-  if (_.size(safariSettings) > 0) {
-    await sim.updateSafariSettings(safariSettings);
-  }
+  return (_.size(safariSettings) > 0)
```
mykola-mokhnach: return !._isEmpty(...) && await sim.updateSafariSettings(...)

lib/settings.js
```diff
@@ -30,8 +30,10 @@ async function setLocaleAndPreferences (sim, opts, safari = false, shutdownFn =
   const localConfig = await setLocale(sim, opts, {}, safari);
   const prefsUpdated = await setPreferences(sim, opts, safari);
   if (localConfig._updated || prefsUpdated) {
-    logger.debug("Updated settings. Rebooting the simulator if it's already open");
+    logger.debug('Updated settings. Rebooting the simulator if it is already open');
     await shutdownFn(sim);
+  } else {
+    logger.debug('Setting did not need to be updated');
```
mykola-mokhnach: no exclamation mark is needed

PR #384, opened 2019-06-20
use isFeatureEnabled for security features check

jlipps:
https://github.com/appium/appium-base-driver/pull/328 needs to be published before this will work

### DIFF
--- a/lib/commands/logging.js
+++ b/lib/commands/logging.js
@@ -3,6 +3,8 @@ import IOSLog from '../device-log/ios-log';
 import IOSCrashLog from '../device-log/ios-crash-log';
 import logger from '../logger';
 
+const GET_SERVER_LOGS_FEATURE = 'get_server_logs';
+
 
 let commands = {}, helpers = {}, extensions = {};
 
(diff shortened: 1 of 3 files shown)


lib/commands/logging.js
```diff
@@ -3,6 +3,8 @@ import IOSLog from '../device-log/ios-log';
 import IOSCrashLog from '../device-log/ios-crash-log';
 import logger from '../logger';
 
+const SERVER_LOGS_FEAT_NAME = 'get_server_logs';
```
mykola-mokhnach: how about `GET_SERVER_LOGS_FEATURE` ?

lib/commands/logging.js
```diff
@@ -38,9 +40,9 @@ extensions.supportedLogTypes = {
   server: {
     description: 'Appium server logs',
     getter: (self) => {
-      if (!self.relaxedSecurityEnabled) {
-        throw new Error('Appium server must have relaxed security flag set ' +
-                        'in order to retrieve server logs');
+      if (!self.isFeatureEnabled(SERVER_LOGS_FEAT_NAME)) {
+        throw new Error(`Security level does not allow feature '${SERVER_LOGS_FEAT_NAME}'. ` +
+                        `See Appium documentation for allowing insecure features.`);
```
imurchie: Can this link to the actual docs?

jlipps: Yes, will work on actual docs. Also will write a function that simply takes a feature name and description and generates this error message, so they all have the same format.

imurchie: That sounds great!

jlipps: `ensureFeatureEnabled` links to actual docs.

PR #400, opened 2019-09-20
fix: better handling page selection on page change in web view

imurchie:
Move setting the context to _before_ setting the page in the remote debugger. This way subsequent calls to functions will get the new context, rather than the old. Since `onPageChange` is called asynchronously, whenever the Web Inspector notifies of a page change, the automating script can be trying to do things with the webview while this change is happening.

In `appium-remote-debugger` setting the page happens immediately but it sometimes takes a while to get a confirmation back from the Web Inspector.

First part of working on https://github.com/appium/appium/issues/13263

### DIFF
--- a/lib/commands/context.js
+++ b/lib/commands/context.js
@@ -363,9 +363,18 @@ extensions.onPageChange = async function onPageChange (pageChangeNotification) {
 
   if (util.hasValue(newPage)) {
     this.selectingNewPage = true;
-    await this.remote.selectPage(appIdKey, parseInt(newPage, 10));
-    this.selectingNewPage = false;
-    this.curContext = `${appIdKey}.${newPage}`;
+    const oldContext = this.curContext;
+    try {
+      this.curContext = `${appIdKey}.${newPage}`;
+      await this.remote.selectPage(appIdKey, parseInt(newPage, 10));
+    } catch (err) {
+      // if there was a problem, do not stay in the new context, which we did
+      // not successfully get into in the Web Inspector
+      this.curContext = oldContext;
+      throw err;
+    } finally {
+      this.selectingNewPage = false;
+    }
   }
   this.windowHandleCache = pageArray;
 };


lib/commands/context.js
```diff
@@ -363,9 +363,18 @@ extensions.onPageChange = async function onPageChange (pageChangeNotification) {
...
-    this.selectingNewPage = false;
-    this.curContext = `${appIdKey}.${newPage}`;
+    const oldContext = this.curContext;
+    try {
+      this.curContext = `${appIdKey}.${newPage}`;
+      await this.remote.selectPage(appIdKey, parseInt(newPage, 10));
+    } catch (err) {
+      // if there was a problem, do not stay in the new context, which we did
+      // not successfully get into in the Web Inspector
+      this.curContext = oldContext;
```
mykola-mokhnach: cannot we simply set the new value to curContext after the call to selectPage is completed? Then it won't be necessary to reset it back here. Or it is used somewhere inside selectPage call?

imurchie: It is used when doing anything else in a webview.

So... say you are on a page and open a new page, to which we automatically switch. The Web Inspector will notify of a new page listing, we will choose it by selecting the page, which notifies `appium-remote-debugger` to talk to it, while also sending a message to the Web Inspector that it is what we want to talk to. The Web Inspector seems to always immediately do that, but does _not_ always reply right away.

If we set the current context _after_ getting a receipt from the Web Inspector, we have to wait as long as that takes in order to communicate with the page, even though we are perfectly capable of doing the communication. So anything the user tries to do goes to the _old_ page while we wait for the Web Inspector to get its act in gear and reply.

imurchie: That being said, I'm still working on getting this tuned right. Closing windows is being annoying.

imurchie:
Moving this into `appium-xcuitest-driver` so as to not mess with this module.

PR #406, opened 2019-12-06
Bump remote debugger to v5

dpgraham:


mykola-mokhnach:
I''m not quite sure the old ios driver supports the recent major changes in the remote debugger

KazuCocoa:
looks good to me, but I would ask @imurchie

imurchie:
This will not work. The `appium-ios-device` package is untested for these versions of iOS. This dependency is locked in Greenkeeper for a reason.

PR #424, opened 2020-03-13
chore: move uuid to appium-support

imurchie:


lib/commands/screenshot.js
```diff
@@ -11,7 +10,7 @@ import { errors } from 'appium-base-driver';
 let commands = {}, helpers = {}, extensions = {};
 
 commands.getScreenshot = async function getScreenshot () {
-  let guid = uuid.create();
+  let guid = util.uuidV4().create();
```
mykola-mokhnach: .create() is not needed.
`util.uuidV4()` already returns a string: https://www.npmjs.com/package/uuid

